PR #4, opened 2023-11-22
TSK-139 create docker-compose.yaml for users-manager

adsukharev:


user-manager/.docker/docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+version: "3.9"
+services:
+  postgres:
+    image: postgres:16.0-alpine
+    container_name: postgres
+    environment:
+      POSTGRES_DB: "user-manager"
```
konstantinwas:
```suggestion
      POSTGRES_DB: "user_manager"
```

https://github.com/RootSoft/Database-Naming-Convention

kostya4201337 (approved):

user-manager/.docker/docker-compose.yaml
adsukharev: good article. 
To not overcomplicate by having lots of docker containers for each service, I propose to name database as `qpay` but each microservice will have dedicated schema like `user_manager` and so on.

Hoocki (approved):

PR #8, opened 2023-12-06
[TSK-141] create Customer model and add migration

Hoocki:


user-manager/src/main/java/com/qpay/usermanager/model/entity/customer/CustomerEntity.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import jakarta.persistence.Entity;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor(access = AccessLevel.PROTECTED)
+@AllArgsConstructor(access = AccessLevel.PROTECTED)
```
adsukharev, line 19: why do you need access level?

user-manager/src/main/resources/application.yaml
```diff
@@ -0,0 +1,15 @@
...
+    schemas: users
+  jpa:
+    show-sql: true
+    properties:
+      hibernate:
+        dialect: org.hibernate.dialect.PostgreSQLDialect
+        format_sql: true
+        default_schema: users
+    hibernate:
+      ddl-auto: update
```
adsukharev: you use flyway for migration, so you don't need to migrate data again using hibernate.
Advantage of flyway is to store logs of migrations and if we something bad happened we can rollback any time. There's a special table in your db "flyway history" as I remember. 
Right now, you use both flyway and hibernate. Instead, it's better to use hibernate ddl to validate correctness of schema `spring.jpa.hibernate.ddl-auto=validate`
https://rieckpil.de/howto-best-practices-for-flyway-and-hibernate-with-spring-boot/
https://www.baeldung.com/database-migrations-with-flyway

user-manager/src/main/resources/application.yaml
```diff
@@ -0,0 +1,15 @@
...
+server:
+  error:
+    include-message: always
```
adsukharev: I don't think it's a good idea to log everything in prod. But it's worth to do it locally. So, move this config to local one.
https://dev.to/flippedcoding/difference-between-development-stage-and-production-d0p

user-manager/src/main/resources/db/migration/V1_create_customers.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS customers
+(
+    id          BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    name        VARCHAR(32),
+    email       VARCHAR(32),
+    password    VARCHAR(32),
+    wallet_id   BIGINT
```
adsukharev: It's important to audit your data. Add created/updatedAt

user-manager/src/main/resources/application-DEV.yaml
adsukharev, line 1: Right now, you manually add DEV profile to you Intellij, correct?
It means, that all developers should do the same thing. It would be great if you can automate this process.
Create folder `.run` in the root and using edit configuration tab create intellij config. If you have problems to create, let's have a chat tomorrow
Screenshot 2023-12-06 at 22 38 04

user-manager/src/main/java/com/qpay/usermanager/model/entity/customer/CustomerEntity.java
Hoocki: i thought, it would be good if constructors have scope only in package or in inheritance classes, because creating instances of this class have to be done by @Builder

adsukharev: never seen it in prod code. However, if you think it's a good idea, I don't mind.

user-manager/src/main/java/com/qpay/usermanager/model/entity/customer/CustomerEntity.java
```diff
@@ -0,0 +1,33 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
+
+    private String name;
+
+    private String email;
+
+    private String password;
+
+    private long walletId;
```
adsukharev, line 35: add create/updated at

user-manager/src/main/resources/db/migration/V001__CREATE_CUSTOMERS.sql
adsukharev, line 1: V1.0__CREATE_CUSTOMERS_TABLE.sql.
Of course, after this change you need to clear your flyway history table to run it again

user-manager/build.gradle.kts
```diff
@@ -19,6 +19,9 @@ dependencies {
 	annotationProcessor("org.projectlombok:lombok")
 
 	implementation("org.springframework.boot:spring-boot-starter-web")
+	implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+	implementation("org.flywaydb:flyway-core:9.22.2")
+	implementation("org.postgresql:postgresql:42.6.0")
```
adsukharev, line 24: Probably lots of your microservices will use similar dependencies with target version. Imagine, later, you are going to update postgres library to 43.0.1 then you need to update this version everywhere.
Instead, you can store versions for all dependencies in variables.

create a new folder `buildSrc` then as for any other project create tree structure: `src/main/{programming language}`
In your case, instead of `java` folder as usual you need to create folder `kotlin` since `buildSrc`  is for gradle and for gradle you use kotlin. Then inside create a file  `Versions.kt`. In this file create an object  with version using kotlin syntax, smth like this:
```
object Versions {
 const val lombok = "1.18.26"
...
}
```
and use it like this:
`compileOnly("org.projectlombok:lombok:${Versions.lombok}")`

https://docs.gradle.org/current/userguide/organizing_gradle_projects.html

adsukharev:
@Hoocki @kostya4201337 
As usual, for each PR reviewer should run the code locally and test it.
Validate that migration works correctly, controllers handle request and send correct response, business logic is valid, UI is responsive, looks nice, errors are handled if you try to use system non standard way and so on. Try to find bugs, break/hack the system. It's an essential skill to find vulnerabilities in the any system.

As a reminder how to review: https://www.notion.so/Code-Review-c7a049b3b5be4bf6a0a368bec41690a3

user-manager/build.gradle.kts
adsukharev, line 1: Let's use newer version of java 21.
https://habr.com/ru/articles/762084/
Don't forget to add new SDK to intellij

adsukharev:
also remove old branches and remove branches automatically after merging

user-manager/src/main/resources/application.yaml
```diff
@@ -0,0 +1,11 @@
+spring:
+  flyway:
+    schemas: users
+  jpa:
+    show-sql: true
```
adsukharev: move `show-sql: true` to local config since logging of sql queries can significantly reduce performance of app on prod

user-manager/src/main/resources/db/migration/V001__CREATE_CUSTOMERS.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS customers
+(
+    id          BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    name        VARCHAR(32),
+    email       VARCHAR(32),
+    password    VARCHAR(32),
+    wallet_id   BIGINT,
+    created_at TIMESTAMP(6),
+    updated_at TIMESTAMP(6)
+);
```
adsukharev: format this file

adsukharev (commented):
remove application.properties file

kostya4201337 (approved):

adsukharev (approved):

PR #9, opened 2023-12-14
[TSK-144] create merchants models and add migration

kostya4201337:


user-manager/src/main/java/com/qpay/usermanager/model/entity/merchant/MerchantEntity.java
```diff
@@ -0,0 +1,42 @@
...
+import lombok.Data;
+import jakarta.persistence.Entity;
+import lombok.NoArgsConstructor;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
+import java.time.LocalDateTime;
+
+@Data
+@NoArgsConstructor(access = AccessLevel.PROTECTED)
+@AllArgsConstructor(access = AccessLevel.PROTECTED)
```
Hoocki, line 19: why u use AccessLevel.PROTECTED?

user-manager/src/main/resources/db/migration/V2.0__CREATE_MERCHANTS_TABLE.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS merchants
+(
+    id         BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    name       VARCHAR(32),
+    email      VARCHAR(32),
+    password   VARCHAR(32),
+    wallet_id  BIGINT,
```
Hoocki: email and wallet_id have to be unique

user-manager/src/main/java/com/qpay/usermanager/model/entity/merchant/MerchantEntity.java
konstantinwas: as you said, 

> it would be good if constructors have scope only in package or in inheritance classes, because creating instances of this class have to be done by @builder

adsukharev (approved):
you are a supreme master of copy/paste

Hoocki (approved):

PR #14, opened 2024-01-04
[TSK-179] create model WalletEntity and migration script

Hoocki:


payment-manager/src/test/java/com/qpay/paymentmanager/repository/WalletRepositoryTest.java
```diff
@@ -0,0 +1,69 @@
...
+        );
+        var wallet2 = testEntityManager.persistAndFlush(
+                WalletEntity.builder()
+                        .balance(BigDecimal.valueOf(500))
+                        .name("wallet2")
+                        .userId(userId)
+                        .userType(userType)
+                        .build()
+        );
+        var wallet3 = testEntityManager.persistAndFlush(
```
konstantinwas:
```suggestion
        testEntityManager.persistAndFlush(
```

you dont have to assign it to the var, you dont use it anyway

payment-manager/src/test/java/com/qpay/paymentmanager/repository/WalletRepositoryTest.java
```diff
@@ -0,0 +1,69 @@
...
+                WalletEntity.builder()
+                        .balance(BigDecimal.valueOf(2000))
+                        .name("wallet3")
+                        .userId(userId)
+                        .userType(UserType.MERCHANT)
+                        .build()
+        );
+        var wallet4 = testEntityManager.persistAndFlush(
```
konstantinwas: same

kostya4201337 (approved):

payment-manager/src/main/java/com/qpay/paymentmanager/repository/WalletRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import com.qpay.paymentmanager.model.entity.wallet.WalletEntity;
+import com.qpay.paymentmanager.model.usertype.UserType;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+import java.util.List;
+
+@Repository
+public interface WalletRepository extends JpaRepository<WalletEntity, Long> {
+
+    List<WalletEntity> findAllByUserIdAndUserType(long userId, UserType userType);
```
adsukharev: We don't support multiple wallets

payment-manager/src/main/java/com/qpay/paymentmanager/model/usertype/UserType.java
```diff
@@ -0,0 +1,6 @@
+package com.qpay.paymentmanager.model.usertype;
+
+public enum UserType {
```
adsukharev: this Enum is common for lots of microservices. What you need to do is to create a new project `libs` and move common models there. For now only this class. Then import this project in your local `build gradle` as a dependency.

payment-manager/src/test/java/com/qpay/paymentmanager/repository/WalletRepositoryTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase;
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
+import org.springframework.boot.test.autoconfigure.orm.jpa.TestEntityManager;
+import org.springframework.test.context.ActiveProfiles;
+import java.math.BigDecimal;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@DataJpaTest()
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@ActiveProfiles(value = "test")
```
adsukharev, line 16: As I remember you don't need to explicitly set profile to `test` in tests. It runs `test` env by default

Hoocki: without it, test fails for reason: missing table [wallets], so i think when you create `application-NAME.yaml` you have to activate profile otherwise it wouldn't use it

payment-manager/src/main/java/com/qpay/paymentmanager/model/entity/wallet/WalletEntity.java
```diff
@@ -0,0 +1,47 @@
+package com.qpay.paymentmanager.model.entity.wallet;
```
adsukharev:
```suggestion
package com.qpay.paymentmanager.model.entity;
```

settings.gradle.kts
```diff
@@ -2,3 +2,5 @@ rootProject.name = "qpay"
 
 include("user-manager")
 include("payment-manager")
+include(":libs")
+project(":libs").projectDir = File(settingsDir, "../libs")
```
adsukharev: what is this line about ? and where is actual `libs` project?

payment-manager/src/main/java/com/qpay/paymentmanager/model/entity/wallet/WalletEntity.java
```diff
@@ -0,0 +1,47 @@
...
+    private long id;
+
+    private String name;
+
+    private BigDecimal balance;
+
+    private long userId;
+
+    @Enumerated(EnumType.STRING)
+    private UserType userType;
```
adsukharev: how do you use `UserType` if it does not exist?

libs/build.gradle.kts
```diff
@@ -0,0 +1,6 @@
+plugins {
+    id("java")
+}
+
+group = "org.qpay"
+version = "unspecified"
```
adsukharev: ?

libs/src/main/java/org/qpay/models/UserType.java
```diff
@@ -0,0 +1,6 @@
+package org.qpay.models;
```
adsukharev:
```suggestion
package com.qpay.libs.models;
```

payment-manager/src/main/java/com/qpay/paymentmanager/model/entity/wallet/WalletEntity.java
Hoocki: i thought `libs` project should be parallel to qpay and obviously it worked only in my PC, so i moved this `libs` into qpay

libs/build.gradle.kts
Hoocki: it's automatically generated, so i thought it needs to determine address of module

adsukharev: Get it back and set version

adsukharev (approved):

PR #15, opened 2024-01-05
[TSK-180] implement crud, add tests

kostya4201337:


payment-manager/src/main/java/com/qpay/paymentmanager/controller/WalletController.java
```diff
@@ -0,0 +1,46 @@
...
+
+@RestController
+@RequiredArgsConstructor
+@RequestMapping(path = PathUtils.WALLET_PATH)
+public class WalletController {
+
+    private final WalletService walletService;
+
+    @GetMapping("{id}")
+    public WalletEntity getWalletById(@PathVariable final Long id) {
```
adsukharev: Why do you use wrapper instead of primitive type?

payment-manager/src/main/java/com/qpay/paymentmanager/model/dto/WalletModification.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.qpay.libs.models.UserType;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import lombok.Builder;
+
+import java.math.BigDecimal;
+
+@Builder
+public record WalletModification (
```
adsukharev, line 11: It is incorrect to use common entity for creation and deletion, because you need different fields during these 2 actions

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+        return walletRepository.save(walletEntity);
+    }
+
+    public WalletEntity updateWallet(WalletModification walletModification, Long id) {
+        var walletEntity = getWalletById(id);
+        walletEntity.setName(walletModification.name());
+        walletEntity.setBalance(walletModification.balance());
+        walletEntity.setUserId(walletModification.userId());
+        walletEntity.setUserType(walletModification.userType());
```
adsukharev: I think it's weird to update user type. Also think about other fields

payment-manager/src/test/java/com/qpay/paymentmanager/controller/WalletControllerIntegrationTest.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    @Test
+    void should_returnException_when_newWalletHasInvalidInput() throws Exception {
+        // given
+        var invalidWallet = WalletModification
+                .builder()
+                .name("")
+                .balance(null)
+                .userId(null)
```
adsukharev: Instead of setting null, don't use these fields in builder

adsukharev:
Task id in commits and branch does not match

payment-manager/src/main/java/com/qpay/paymentmanager/model/dto/WalletModification.java
konstantinwas: but i dont use walletModification for deletion

adsukharev: I meant creation and modification

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+
+        return walletRepository.save(walletEntity);
+    }
+
+    public WalletEntity updateWallet(WalletModification walletModification, long id) {
+        var walletEntity = getWalletById(id);
+        walletEntity.setName(walletModification.name());
+        walletEntity.setBalance(walletModification.balance());
+        walletEntity.setUserId(walletModification.userId());
+
```
adsukharev: Empty line

payment-manager/src/main/java/com/qpay/paymentmanager/controller/WalletController.java
adsukharev: You did not answer the question

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
adsukharev: What about other fields ?

payment-manager/src/main/java/com/qpay/paymentmanager/controller/WalletController.java
konstantinwas: i thought long doesnt have primitive type as well as string for some reason

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
```diff
@@ -21,18 +22,14 @@ public WalletEntity getWalletById(long id) {
         return walletRepository.findById(id).orElseThrow(NoWalletFoundException::new);
     }
 
-    public WalletEntity addWallet(WalletModification walletModification) {
-        final var walletEntity = walletMapper.map(walletModification);
-
+    public WalletEntity addWallet(WalletCreation walletCreation) {
```
adsukharev: Did you set up `final` in your Intellij ? There are lots of places where you miss it

payment-manager/src/test/java/com/qpay/paymentmanager/controller/WalletControllerIntegrationTest.java
adsukharev: in your case, you can create just empty entity: `WalletModification.builder().build();`

adsukharev:
I checked other projects, we have to change a bit `build.gradle files.`
1) remove group and version from local `build.gradle` files in your sub projects
2) add version to root `build.gralde`

payment-manager/src/main/java/com/qpay/paymentmanager/mapper/WalletMapper.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.math.BigDecimal;
+
+@Component
+public class WalletMapper {
+
+    public WalletEntity map(final WalletCreation walletCreation) {
+        return WalletEntity.builder()
+                .name(walletCreation.name())
+                .balance(new BigDecimal(0))
```
adsukharev: instead of setting default value over here and in some places in your tests. You can add default value using builder for your walletEntity class https://www.baeldung.com/lombok-builder-default-value#builder-default

payment-manager/src/main/java/com/qpay/paymentmanager/utility/PathUtils.java
```diff
@@ -0,0 +1,9 @@
+package com.qpay.paymentmanager.utility;
```
adsukharev:
```suggestion
package com.qpay.paymentmanager.utils;
```

adsukharev:
mover `server.port` config to application-DEV from all projects

adsukharev (approved):

payment-manager/src/main/java/com/qpay/paymentmanager/utils/PathUtils.java
Hoocki, line 1: I think you could move `PathUtils` into `libs` module, because we duplicate this class in every module

adsukharev: It's better to keep a dedicated class for each project

Hoocki (approved):

PR #17, opened 2024-01-08
[TSK-184] add getWalletByUser method

kostya4201337:


payment-manager/src/main/java/com/qpay/paymentmanager/controller/WalletController.java
```diff
@@ -28,6 +22,11 @@ public WalletEntity getWalletById(@PathVariable final long id) {
         return walletService.getWalletById(id);
     }
 
+    @GetMapping("user/{userId}")
```
adsukharev: All paths should be in plural form

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
```diff
@@ -22,6 +23,14 @@ public WalletEntity getWalletById(final long id) {
         return walletRepository.findById(id).orElseThrow(NoWalletFoundException::new);
     }
 
+    public WalletEntity getWalletByUser(final long userId, final UserType userType) {
+        final WalletEntity walletEntity = walletRepository.findWalletEntityByUserIdAndUserType(userId, userType);
+        if (walletEntity == null) {
+            throw new NoWalletFoundException();
+        }
```
adsukharev, line 30: Use the same approach of checking null as you did in another method to get by id

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
```diff
@@ -22,6 +23,14 @@ public WalletEntity getWalletById(final long id) {
         return walletRepository.findById(id).orElseThrow(NoWalletFoundException::new);
     }
 
+    public WalletEntity getWalletByUser(final long userId, final UserType userType) {
+        final WalletEntity walletEntity = walletRepository.findWalletEntityByUserIdAndUserType(userId, userType);
```
adsukharev: Var

adsukharev: If this query is going to be used often, it can be a bottleneck. Can you explain why, and how can we solve this issue ?

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
konstantinwas: image

i cant

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
konstantinwas: beacuse searching for wallet by 2 vars takes long time

adsukharev: what is the difference between searching by id and any other field ?

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
adsukharev: make a repository method return type `optional`

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/WalletServiceImpl.java
konstantinwas: search by ID performs instantly

adsukharev: why ?

konstantinwas: because it is primary key

adsukharev: Why we can search fast by primary key?

konstantinwas: because it's like index in array

adsukharev: What can we do to increase reading speed of other fields ?

adsukharev: @kostya4201337 do you understand what you need to do?

konstantinwas: create index

adsukharev: For which field/fields ?

konstantinwas: user id and user type

adsukharev: You are so fast to answer. 

Let's think a bit. When we use index we reduce speed of writing because we need to add this data structure in addition to actual data. Can we improve your idea and use only one index?

For each user id it's possible to have max 2 elements. But for each user type you can have lots of elements. Based on it can we use only one field for index ? And which one?

konstantinwas: we need to create index for user id because 

> For each it's possible to have max 2 elements.

payment-manager/src/main/resources/db/migration/V1.0__CREATE_WALLETS_TABLE.sql
```diff
@@ -7,4 +7,6 @@ CREATE TABLE IF NOT EXISTS wallets
     user_type  VARCHAR(32),
     created_at TIMESTAMP(6),
     updated_at TIMESTAMP(6)
-);
\ No newline at end of file
+);
+
+CREATE INDEX IF NOT EXISTS user_id_index ON wallets(user_id);
```
adsukharev, line 12: Actually this script has been merged. it's kind of in "production". So, you cannot rewrite this script, you can only create a new one. But this time, ok, you can leave as it is

adsukharev (approved):

payment-manager/src/main/resources/db/migration/V1.0__CREATE_WALLETS_TABLE.sql
konstantinwas: should i delete this branch since we don't need it?

adsukharev: Do you mean the whole feature in general ?

konstantinwas: getWalletByUser method

adsukharev: As you've already done it, merge it

Hoocki (approved):

PR #18, opened 2024-01-19
[TSK-183] add request to transaction history in makePayment method

kostya4201337:


payment-manager/src/main/java/com/qpay/paymentmanager/PaymentManagerApplication.java
```diff
@@ -10,4 +12,8 @@ public static void main(final String[] args) {
         SpringApplication.run(PaymentManagerApplication.class, args);
     }
 
+    @Bean
+    public RestTemplate restTemplate(){
+        return new RestTemplate();
+    }
```
adsukharev: All beans should be created in config files, see https://www.geeksforgeeks.org/spring-configuration-annotation-with-example/
remember SRP (single responsibility principle)? Our main class is only responsible to run the app.

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
```diff
@@ -33,6 +42,20 @@ public WalletEntity makePayment(final WalletPayment walletPayment) {
         final var updatedToWalletBalance = toWallet.getBalance().add(walletPayment.amount());
         updateBalance(updatedToWalletBalance, toWallet);
 
+        final HttpHeaders httpHeaders = new HttpHeaders();
```
adsukharev: var

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
```diff
@@ -33,6 +42,20 @@ public WalletEntity makePayment(final WalletPayment walletPayment) {
...
+
+        final Map<String, String> requestBody = new HashMap<>();
+        requestBody.put("name_from", fromWallet.getName());
+        requestBody.put("id_from", String.valueOf(walletPayment.walletIdFrom()));
+        requestBody.put("name_to", toWallet.getName());
+        requestBody.put("id_to", String.valueOf(walletPayment.walletIdTo()));
+        requestBody.put("amount", walletPayment.amount().toString());
+
+        final HttpEntity<Map<String, String>> requestEntity = new HttpEntity<>(requestBody, httpHeaders);
+
```
adsukharev: no need to add indents everywhere. When you do it, it means logic of the method is big and it's hard to read. In such cases, consider divide  it into several methods or classes

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
```diff
@@ -33,6 +42,20 @@ public WalletEntity makePayment(final WalletPayment walletPayment) {
...
+        restTemplate.postForEntity(PathUtils.TRANSACTION_HISTORY_PATH, requestEntity, String.class);
```
adsukharev: why did you choose String as a return type?

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
```diff
@@ -33,6 +42,20 @@ public WalletEntity makePayment(final WalletPayment walletPayment) {
...
 
+        final HttpHeaders httpHeaders = new HttpHeaders();
+        httpHeaders.setContentType(MediaType.APPLICATION_JSON);
+
+        final Map<String, String> requestBody = new HashMap<>();
+        requestBody.put("name_from", fromWallet.getName());
+        requestBody.put("id_from", String.valueOf(walletPayment.walletIdFrom()));
+        requestBody.put("name_to", toWallet.getName());
+        requestBody.put("id_to", String.valueOf(walletPayment.walletIdTo()));
+        requestBody.put("amount", walletPayment.amount().toString());
```
adsukharev: I would be really lazy to manually add each field. 
https://www.baeldung.com/rest-template#exchange_post

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
adsukharev: Use more generic `exchange`
https://www.baeldung.com/rest-template#exchange_post

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
```diff
@@ -33,6 +42,20 @@ public WalletEntity makePayment(final WalletPayment walletPayment) {
         final var updatedToWalletBalance = toWallet.getBalance().add(walletPayment.amount());
         updateBalance(updatedToWalletBalance, toWallet);
 
+        final HttpHeaders httpHeaders = new HttpHeaders();
+        httpHeaders.setContentType(MediaType.APPLICATION_JSON);
```
adsukharev: I'm not sure, that we need to explicitly add this header

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
```diff
@@ -33,6 +42,20 @@ public WalletEntity makePayment(final WalletPayment walletPayment) {
...
+        requestBody.put("name_from", fromWallet.getName());
+        requestBody.put("id_from", String.valueOf(walletPayment.walletIdFrom()));
+        requestBody.put("name_to", toWallet.getName());
+        requestBody.put("id_to", String.valueOf(walletPayment.walletIdTo()));
+        requestBody.put("amount", walletPayment.amount().toString());
+
+        final HttpEntity<Map<String, String>> requestEntity = new HttpEntity<>(requestBody, httpHeaders);
+
+        restTemplate.postForEntity(PathUtils.TRANSACTION_HISTORY_PATH, requestEntity, String.class);
+
```
adsukharev, line 52: You have to extract logic of sending data to  `client/TransactionHistoryClient.java`

payment-manager/src/main/java/com/qpay/paymentmanager/utils/PathUtils.java
```diff
@@ -8,4 +8,5 @@ public class PathUtils {
 
     public static final String PAYMENT_PATH = "/api/v1/payments";
 
+    public static final String TRANSACTION_HISTORY_PATH = "api/v1/history";
```
adsukharev: This class is for controller. Move it to a newly created client class.

adsukharev (commented):
You have to add integration tests to check that your client works. Right now, it's pretty clear that it does not work. And you even did not test it yourself. 

It's clear that you did not test it, since you did not set host of the target endpoint (where you send a request).
You send a request to `api/transactions...` but it's just a subdomain. What is the host ? 
`http://api/transactions` or `http://google.com/api/transactions` or anything else?

Then think about different environments. For your local development it can be one URL, but for production another.

For example for local: http://123.213.123/api/transactions
but for prod: http://qpay.com/api/transactions

adsukharev:
Don't forget to ask questions when you are stuck

adsukharev:
Roma said `restTemplate` is deprecated. Use webclient

As of Spring Framework 5, alongside the WebFlux stack, Spring introduced a new HTTP client called [WebClient](https://www.baeldung.com/spring-5-webclient).

WebClient is a modern, alternative HTTP client to [RestTemplate](https://www.baeldung.com/spring-webclient-resttemplate). Not only does it provide a traditional synchronous API, but it also supports an efficient nonblocking and asynchronous approach.

That said, if we’re developing new applications or migrating an old one, it’s a good idea to use WebClient. Moving forward, RestTemplate will be deprecated in future versions.

payment-manager/src/main/java/com/qpay/paymentmanager/model/dto/TransactionCreation.java
```diff
@@ -0,0 +1,20 @@
+package com.qpay.paymentmanager.model.dto;
+
+import lombok.Builder;
+
+import java.math.BigDecimal;
+
+@Builder
+public record TransactionCreation (
```
adsukharev:
```suggestion
public record PaymentTransaction (
```

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
```diff
@@ -40,6 +44,14 @@ public WalletEntity makePayment(final WalletPayment walletPayment) {
         updateBalance(updatedToWalletBalance, toWallet);
 
         executorService.execute(() -> notificationClient.sendNotification(walletPayment));
+        transactionHistoryClient.saveTransactionToHistory(new TransactionCreation(
```
adsukharev: create data using builder in a dedicated variable.
Also move this data and request itself to a separate method

payment-manager/src/main/resources/application.yaml
```diff
@@ -1,9 +1,9 @@
 webclient:
   services:
     notification-manager:
-      base-url: http://localhost:8083
-    transaction-manager:
-      base-url: http://localhost:8084
+      base-url: http://localhost:8183
+    transaction-history-manager:
+      base-url: http://localhost:8184
```
adsukharev: leave ports as it was. We use 81.. for mock servers

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
```diff
@@ -40,6 +44,14 @@ public WalletEntity makePayment(final WalletPayment walletPayment) {
         updateBalance(updatedToWalletBalance, toWallet);
 
         executorService.execute(() -> notificationClient.sendNotification(walletPayment));
+        saveTransactionToHistory(
+                fromWallet.getName(),
+                toWallet.getName(),
+                walletPayment.walletIdFrom(),
+                walletPayment.walletIdTo(),
+                walletPayment.amount()
```
adsukharev: you can pass ` walletPayment` as one parameter. Then get all necessary fields in `saveTransactionToHistory` method

adsukharev (approved):

payment-manager/src/main/java/com/qpay/paymentmanager/service/impl/PaymentServiceImpl.java
konstantinwas: i cant get names from walletPayment

adsukharev: But you can get wallet ids and amount

Hoocki (approved):

PR #20, opened 2024-01-26
[TSK-197] init notification-manager

Hoocki:


notification-manager/README.md
```diff
@@ -0,0 +1,2 @@
+# Notification manager
+Service for sends information about transaction to email
```
adsukharev:
```suggestion
Service for sending emails about payments
```

notification-manager/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
+plugins {
+	java
+	id("org.springframework.boot") version "3.2.1"
```
adsukharev:
```suggestion
	id("org.springframework.boot") version "3.2.2"
```

notification-manager/src/main/resources/application.yaml
```diff
@@ -0,0 +1,6 @@
+spring:
+  application:
+    name: notification-manager
+
+server:
+  port: 8083
```
adsukharev: this is for DEV. Check other services, if this config in default one, move it to DEV as well

adsukharev (approved):

PR #26, opened 2024-02-14
[TSK-176] implement qr generation, add tests

kostya4201337:


qr-generator/src/main/java/com/qpay/qrgenerator/controller/QrGeneratorController.java
```diff
@@ -0,0 +1,28 @@
...
+
+@RestController
+@RequiredArgsConstructor
+@RequestMapping(path = PathUtils.QR_GENERATOR_PATH)
+public class QrGeneratorController {
+
+    private final QrGeneratorService qrGeneratorService;
+
+    @PostMapping(
+            value = "/generate",
```
adsukharev: you can extract this path as well.
`generate` is a bad path name since it is a verb.
I would just remove it at all

qr-generator/src/main/java/com/qpay/qrgenerator/utils/PathUtils.java
```diff
@@ -0,0 +1,9 @@
+package com.qpay.qrgenerator.utils;
+
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class PathUtils {
+    public static final String QR_GENERATOR_PATH = "/api/v1/qr";
```
adsukharev:
```suggestion
    public static final String QR_GENERATOR_BASE_PATH = "/api/v1/qrcode";
```

qr-generator/src/main/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class QrGeneratorServiceImpl implements QrGeneratorService {
+
+    private static final int QR_CODE_SIZE = 240;
+
+    private final ObjectMapper objectMapper;
+
+    public byte[] generateQrCode(final QrCodeCreation qrCodeCreation) throws IOException, WriterException {
```
adsukharev: handle exception in this method and throw your custom

qr-generator/src/main/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import com.qpay.qrgenerator.service.QrGeneratorService;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+@Service
+@RequiredArgsConstructor
+public class QrGeneratorServiceImpl implements QrGeneratorService {
```
adsukharev: `Impl` is a default name when you don't have imagination. 

As you know we can have lots of implementations for one interface.
Right now your implementation is base on `zxing` library. What if in the future we want to have another implementation for other cases?

If you look at the notification manager, there's a `NotificationService` interface. Straightforward approach is to name  `NotificationServiceImpl`. However, we can do many different implementations ` EmailNotification`, `TelegramNotification` and so on. In our case Roma named implementation as `EmailNotification`.

In general, it's ok to put `impl` in the end, but if we have concrete implementation let's name it more precise.
In your case, I would call it `QrCodeGeneratorZxing`

qr-generator/src/main/java/com/qpay/qrgenerator/model/dto/QrCodeCreation.java
```diff
@@ -0,0 +1,21 @@
...
+import lombok.Builder;
+
+import java.math.BigDecimal;
+
+@Builder
+public record QrCodeCreation (
+        long walletId,
+
+        @NotBlank(message = "Username is required")
+        String userName,
```
adsukharev:
```suggestion
        String username,
```

adsukharev: what is username ? For our users we have only email and name fields

qr-generator/src/main/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private final ObjectMapper objectMapper;
+
+    public byte[] generateQrCode(final QrCodeCreation qrCodeCreation) throws IOException, WriterException {
+        final var json = objectMapper.writeValueAsString(qrCodeCreation);
+
+        final var qrCodeWriter = new QRCodeWriter();
+        final var outputStream = new ByteArrayOutputStream();
+        final var bitMatrix = qrCodeWriter.encode(json, BarcodeFormat.QR_CODE, QR_CODE_SIZE, QR_CODE_SIZE);
+        MatrixToImageWriter.writeToStream(bitMatrix, "PNG", outputStream, new MatrixToImageConfig(MatrixToImageConfig.WHITE, MatrixToImageConfig.BLACK));
```
adsukharev: move png to constant

qr-generator/src/main/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
```
adsukharev: remove empty line

qr-generator/src/main/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+@RequiredArgsConstructor
+public class QrGeneratorServiceImpl implements QrGeneratorService {
+
+    private static final int QR_CODE_SIZE = 240;
+
+    private final ObjectMapper objectMapper;
+
+    public byte[] generateQrCode(final QrCodeCreation qrCodeCreation) throws IOException, WriterException {
+        final var json = objectMapper.writeValueAsString(qrCodeCreation);
+
```
adsukharev: remove empty line

qr-generator/src/test/java/com/qpay/qrgenerator/controller/QrGeneratorControllerIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @MockBean
+    private QrGeneratorServiceImpl qrGeneratorService;
+
+    private static final int QR_CODE_SIZE = 240;
+
+    @Test
+    void should_returnQrCode() throws Exception {
+        //given
+        var qrCodeCreation = QrCodeCreation.builder().walletId(1).userName("bob").money(new BigDecimal(1234)).build();
```
adsukharev: don't write everything in one line

qr-generator/src/test/java/com/qpay/qrgenerator/controller/QrGeneratorControllerIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @MockBean
+    private QrGeneratorServiceImpl qrGeneratorService;
+
+    private static final int QR_CODE_SIZE = 240;
+
+    @Test
+    void should_returnQrCode() throws Exception {
```
adsukharev, line 33: This test is weird since you exactly implement your service class here. We all know you are a **_legend_** of copy/paste but what if a service contains hundreds of lines, will you copy it all? Then somebody will change a bit code, then you will have to copy/paste again.

What you are checking here is that controller returns array of bytes. It does not matter content of it.
In your `given` stub you can return any array of byte.

qr-generator/src/test/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+
+import java.io.ByteArrayOutputStream;
+import java.math.BigDecimal;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SpringBootTest
```
adsukharev: do you know what this annotation do?

qr-generator/src/test/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Autowired
+    private QrGeneratorServiceImpl qrGeneratorService;
+
+    @Autowired
+    ObjectMapper objectMapper;
+
+    private static final int QR_CODE_SIZE = 240;
+
+    @Test
+    void should_GenerateQrCode() throws Exception {
```
adsukharev:
```suggestion
    void should_generateQrCode() throws Exception {
```

qr-generator/src/test/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImplTest.java
```diff
@@ -0,0 +1,46 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SpringBootTest
+class QrGeneratorServiceImplTest {
+
+    @Autowired
+    private QrGeneratorServiceImpl qrGeneratorService;
+
+    @Autowired
+    ObjectMapper objectMapper;
```
adsukharev:
```suggestion
    private ObjectMapper objectMapper;
```

qr-generator/src/test/java/com/qpay/qrgenerator/service/impl/QrGeneratorServiceImplTest.java
adsukharev: First of all, it is not integration test. This is a unit test where you check only this class. You don't need to start the whole app.

In general, test is useless.
You do not need to repeat your code. What you are doing now is testing qr code library. But it does not make sense. It's already tested by developers of the library.

You can either check that result of this call `qrGeneratorService.generateQrCode(qrCodeCreation);` is not empty or generate beforehand qr code and check it with the result. Second approach is better but it only works if qr codes are same all the time for the same data. If it is not, it won't work.  So, for simplicity check that result is not empty

adsukharev (commented):
generate config file for intellij in .run folder. Otherwise it's not convenient to set env variable manually

adsukharev:
Previous time you merged your PR without approval from Roma. Don't do it next times

qr-generator/src/main/java/com/qpay/qrgenerator/service/impl/QrCodeGeneratorZxing.java
```diff
@@ -0,0 +1,64 @@
...
+    private ByteArrayOutputStream writeQrCodeToImage(final BitMatrix bitMatrix) {
+        final var outputStream = new ByteArrayOutputStream();
+        try {
+            MatrixToImageWriter.writeToStream(bitMatrix, QR_CODE_IMAGE_TYPE, outputStream, new MatrixToImageConfig(MatrixToImageConfig.WHITE, MatrixToImageConfig.BLACK));
+        } catch (final IOException e) {
+            throw new QrCodeToImageWriteException(e.getMessage());
+        }
+        return outputStream;
+    }
+}
```
adsukharev, line 38: Let's make it easier:
```
public byte[] generateQrCode(final QrCodeCreation qrCodeCreation) {
        final var qrCodeWriter = new QRCodeWriter();
        final var outputStream = new ByteArrayOutputStream();
        try {
            final var json = objectMapper.writeValueAsString(qrCodeCreation);
            final var bitMatrix = qrCodeWriter.encode(json, BarcodeFormat.QR_CODE, QR_CODE_SIZE, QR_CODE_SIZE);
            MatrixToImageWriter.writeToStream(bitMatrix, QR_CODE_IMAGE_TYPE, outputStream, new MatrixToImageConfig(MatrixToImageConfig.WHITE, MatrixToImageConfig.BLACK));
        }  catch (final Exception err) {
            throw new QrEncodeException(err.getMessage());
        }
        return outputStream.toByteArray();
    }
```

qr-generator/src/test/java/com/qpay/qrgenerator/controller/QrGeneratorControllerIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+        //given
+        var qrCodeCreation = QrCodeCreation.builder()
+                .walletId(1)
+                .name("bob")
+                .money(new BigDecimal(1234))
+                .build();
+        given(qrGeneratorService.generateQrCode(qrCodeCreation)).willReturn(new byte[1]);
```
adsukharev:
```
var qrCodeBytes = "Qr code data".getBytes();
given(qrGeneratorService.generateQrCode(qrCodeCreation)).willReturn(qrCodeBytes);
```

qr-generator/src/test/java/com/qpay/qrgenerator/controller/QrGeneratorControllerIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+                .perform(MockMvcRequestBuilders
+                        .post(PathUtils.QR_GENERATOR_BASE_PATH + "/qrcode")
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .content(objectMapper.writeValueAsString(qrCodeCreation)))
+                .andReturn()
+                .getResponse()
+                .getContentAsByteArray();
+
+        //then
+        assertThat(responseBody).isInstanceOf(byte[].class);
```
adsukharev: 1.  Of course `isInstance` is `byte[]` since a few lines ago you converted it to bytes. This assertion is useless. It's like create a string `String kek = "cheburek"; ` and check that kek is string.
2. You have to validate exact data you get from the controller:
`assertThat(responseBody).isEqualTo(qrCodeBytes);`
2. Do you find it strange that your response is empty?

qr-generator/src/test/java/com/qpay/qrgenerator/service/impl/QrCodeGeneratorZxingTest.java
```diff
@@ -0,0 +1,30 @@
...
+                .money(new BigDecimal(1234))
+                .build();
+
+        //when
+        var image = qrGeneratorService.generateQrCode(qrCodeCreation);
+
+        //then
+        assertThat(image).isNotEmpty();
+    }
+}
```
adsukharev: Not all lines are tested. It's good to test exception block

adsukharev (approved):

Hoocki (approved):

PR #30, opened 2024-02-17
[TSK-203] update spring boot versions

kostya4201337:


adsukharev (dismissed):

adsukharev:
Do you see how it is annoying to update all versions manually. You can make it smarter. 
If you add plugin section to root build gradle. Like this for all plugins:
```
plugins {
	id("org.springframework.boot") version "3.2.2" apply false
	id("io.spring.dependency-management") version "1.1.4" apply false
//		... add here other plugins from your submodules
}
```
`apply false` means that this plugin is not applied for this project. In our case root. And yes, it would be strange to apply spring boot plugin for root project. But it will be applied for all submodules where you use this plugin

Then for you submodules you can skip version 
```
plugins {
    java
    id("org.springframework.boot")
    id("io.spring.dependency-management")
}
```

See example here https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block

build.gradle.kts
```diff
@@ -20,4 +20,9 @@ allprojects {
 		set("zxingCore", "3.5.3")
 		set("zxingJavaSe", "3.5.3")
 	}
+}
+
+plugins {
+	id("org.springframework.boot") version "3.2.2" apply false
+	id("io.spring.dependency-management") version "1.1.4" apply false
```
adsukharev: move it on top

adsukharev (commented):
Check `libs` project, there's one more plugin

adsukharev (approved):

adsukharev:
Don't wait other approvals, keep working on new tasks and create new PRs

Hoocki (approved):

PR #33, opened 2024-02-18
[TSK-212] change application-DEV to application-LOCAL

Hoocki:


kostya4201337 (approved):

adsukharev (approved):
Sorry, if I said unclear, but preferable way to use `local` is in lowercase. `application-local.yaml`

PR #36, opened 2024-02-20
[TSK-218] initialized auth-manager

Hoocki:


adsukharev (approved):
Since this is just initialization, you can merge it without Kostya's approval

PR #37, opened 2024-02-20
[TSK-216] init transaction history manager

kostya4201337:


konstantinwas:
@adsukharev check design

Hoocki (approved):

transaction-history-manager/src/main/resources/application-local.yaml
```diff
@@ -0,0 +1,4 @@
+server:
+  port: 8090
```
adsukharev, line 2: Use port number that goes next after auth manager

adsukharev (approved):

PR #41, opened 2024-02-24
[TSK-221] implement JWT service

Hoocki:


auth-manager/src/main/java/com/qpay/authmanager/service/authentication/impl/AuthenticationServiceImpl.java
```diff
@@ -18,7 +18,10 @@ public class AuthenticationServiceImpl implements AuthenticationService {
 
     public JwtAuthenticationResponse signIn(final AuthCredentials authCredentials) {
         final var userEntity = userService.getUserByEmail(authCredentials.email());
-        return jwtService.generate(userEntity);
+        final String generatedToken = jwtService.generateToken(userEntity.getEmail());
```
adsukharev: Var

auth-manager/src/main/java/com/qpay/authmanager/service/authentication/impl/AuthenticationServiceImpl.java
```diff
@@ -18,7 +18,10 @@ public class AuthenticationServiceImpl implements AuthenticationService {
...
+        return JwtAuthenticationResponse.builder()
+                .token(generatedToken)
+                .build();
```
adsukharev: I would use constructor here because there is only one value

adsukharev (approved):

PR #48, opened 2024-02-26
[TSK-172] create buildSrc module and move versions

Hoocki:
finally

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,21 +7,6 @@ plugins {
 group = "com.qpay"
 version = "0.0.1-SNAPSHOT"
 
-
-ext {
-	set("flyway", "9.22.2")
-	set("postgresql", "42.6.0")
-	set("junit-jupiter-api", "5.10.0")
-	set("h2", "2.2.224")
-	set("webFlux", "3.2.2")
-	set("wireMock", "3.3.1")
-	set("jakartaValidation", "3.0.0")
-	set("kafka", "3.1.1")
-	set("zxingCore", "3.5.3")
-	set("zxingJavaSe", "3.5.3")
-}
-
-
 tasks.wrapper {
 	gradleVersion = "8.6"
 	distributionType = Wrapper.DistributionType.ALL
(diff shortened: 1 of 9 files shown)


adsukharev:
Resolve conflicts, tomorrow I will check

build.gradle.kts
```diff
@@ -4,24 +4,12 @@ plugins {
 	id("io.freefair.lombok") version "8.4" apply false
 }
 
-group = "com.qpay"
-version = "0.0.1-SNAPSHOT"
```
adsukharev, line 8: why did you remove group and version?

build.gradle.kts
```diff
@@ -4,24 +4,12 @@ plugins {
...
-	set("webFlux", "3.2.2")
-	set("wireMock", "3.3.1")
-	set("jakartaValidation", "3.0.0")
-	set("kafka", "3.1.1")
-	set("zxingCore", "3.5.3")
-	set("zxingJavaSe", "3.5.3")
+allprojects {
+	repositories {
+		mavenCentral()
+	}
```
adsukharev: you incorrectly rebased branch. Check previous changes or ask Kostya

libs/build.gradle.kts
```diff
@@ -13,10 +13,8 @@ configurations {
     }
 }
 
-val jakartaValidation = rootProject.extra["jakartaValidation"]
-
 dependencies {
 
-    implementation("jakarta.validation:jakarta.validation-api:$jakartaValidation")
+    implementation("jakarta.validation:jakarta.validation-api:${Versions.JAKARTA_VALIDATION}")
```
adsukharev, line 18: As I said previously, use direct import

build.gradle.kts
Hoocki: i thought, it's optional and can be removed, but it would be good to know group of project and version, so i will return it

adsukharev: I don't see that you returned it

adsukharev (approved):

PR #54, opened 2024-03-08
[TSK-238] create base plugin

kostya4201337:


buildSrc/src/main/kotlin/com.qpay.java.base.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+configurations {
+    compileOnly {
+        extendsFrom(configurations.annotationProcessor.get())
+    }
+}
+
+dependencies {
+    annotationProcessor("org.projectlombok:lombok")
+
+    implementation("org.springframework.boot:spring-boot-starter-web")
```
adsukharev: This plugin is for java. For spring boot, you can create separate one but no need now.
Also it's not a good idea to add spring-boot-web to base as it makes us use exactly this framework. But probably some of the services will use spring webflux or micronaut.

adsukharev (commented):
remove the `java` plugin from your submodules since it is already in the `tests` plugin

adsukharev (approved):

Hoocki (approved):

PR #57, opened 2024-03-10
[TSK-240] init api-gateway

Hoocki:


api-gateway/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+
+java {
+	sourceCompatibility = JavaVersion.VERSION_21
+}
+
+configurations {
+	compileOnly {
+		extendsFrom(configurations.annotationProcessor.get())
+	}
+}
```
adsukharev: `base` plugin has all related to java

adsukharev (commented):
Let's make this module a bit different.
Create module `cloud`. Inside of it, create a submodule `api-gateway`
Then change in settings gradle from `api-gateway` to `cloud:api-gateway` or `:cloud:api-gateway`.

Later, we will create inside of `cloud` service-discovery

adsukharev (approved):

PR #59, opened 2024-03-11
[TSK-241] configure gateway

Hoocki:


cloud/api-gateway/src/main/resources/application-local.yaml
```diff
@@ -1,4 +1,28 @@
...
+          predicates:
+            - Path=/api/v1/notifications/**
+        - id: qr-generator
+          uri: http://localhost:8085
+          predicates:
+            - Path=/api/v1/qrcode/**
+        - id: auth-manager
+          uri: http://localhost:8086
+          predicates:
+            - Path=/api/v1/users/**,/api/v1/auth/**
```
adsukharev:
```suggestion
            - Path=/api/v1/auth/users/**,/api/v1/auth/**
```
I don't like that for auth manager the path is `/api/v1/users/` .  Looks like it is for user-manager.

cloud/api-gateway/src/main/resources/application-local.yaml
```diff
@@ -1,4 +1,28 @@
 server:
   port: 8080
   error:
-    include-message: always
\ No newline at end of file
+    include-message: always
+spring:
```
adsukharev: move to prod config

cloud/api-gateway/src/main/resources/application-local.yaml
```diff
@@ -1,4 +1,28 @@
...
+        - id: payment-manager
+          uri: http://localhost:8082
+          predicates:
+            - Path=/api/v1/wallets/**,/api/v1/payments/**
+        - id: notification-manager
+          uri: http://localhost:8083
+          predicates:
+            - Path=/api/v1/notifications/**
+        - id: qr-generator
+          uri: http://localhost:8085
```
adsukharev: strange that you don't have 8084 port

adsukharev (commented):
Did you check that it works?
I don't see main file

cloud/api-gateway/src/main/resources/application-local.yaml
Hoocki: 8084 - transaction-history

Hoocki:
> Did you check that it works? I don't see main file

i checked, but i don't understand which main file is it about

adsukharev:
> > Did you check that it works? I don't see main file
> 
> i checked, but i don't understand which main file is it about

I did not see ApiGatewayApplication. But you have already created it, no problem

adsukharev:
add logger
```
logging:
  level:
    org.springframework.cloud.gateway: DEBUG
    reactor.netty.http.client: DEBUG
```

cloud/api-gateway/src/main/resources/application.yaml
```diff
@@ -1,3 +1,31 @@
...
+            predicates:
+              - Path=/api/v1/notifications/**
+          - id: qr-generator
+            uri: http://localhost:8085
+            predicates:
+              - Path=/api/v1/qrcode/**
+          - id: auth-manager
+            uri: http://localhost:8086
+            predicates:
+              - Path=/api/v1/identity/users/**,/api/v1/auth/**
```
adsukharev: Actually this path /api/v1/identity/users/** should not be in the API Gateway. This is for internal use only. We don't allow users directly send requests to authManager to create users. Same for notifications
Btw identity is not relevant name. Maybe `credentials` ?

cloud/api-gateway/src/main/resources/application.yaml
```diff
@@ -1,3 +1,31 @@
...
+
+logging:
+  level:
+    org.springframework.cloud.gateway: DEBUG
+    reactor.netty.http.client: DEBUG
```
adsukharev: debug logger is for local

cloud/api-gateway/src/main/resources/application.yaml
```diff
@@ -1,3 +1,26 @@
...
+        routes:
+          - id: user-manager
+            uri: http://localhost:8081
+            predicates:
+              - Path=/api/v1/customers/**,/api/v1/merchants/**
+          - id: payment-manager
+            uri: http://localhost:8082
+            predicates:
+              - Path=/api/v1/wallets/**,/api/v1/payments/**
+          - id: notification-manager
```
adsukharev, line 15: for internal use. Remove it

adsukharev (approved):

cloud/api-gateway/src/main/resources/application.yaml
Hoocki: remove notfication service or paths in payment-manager?

Hoocki: i thought these paths should be able for user to get wallet and provide payment and so on

adsukharev: Users cannot directly send requests to notification manager.

PR #62, opened 2024-03-15
[TSK-247] add circuit breaker

Hoocki:


cloud/api-gateway/src/main/resources/application.yaml
```diff
@@ -9,20 +9,62 @@ spring:
...
+    health:
+      show-details: always
+
+resilience4j:
+  circuitbreaker:
+    configs:
+      default:
+        register-health-indicator: true
+        sliding-window-type: TIME_BASED
+        sliding-window-size: 1
```
adsukharev, line 49: Does it mean that even if you get only one failure circuit breaker switches to open state ?

adsukharev (commented):
Could you comment each property what it means and why you set such value

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -14,4 +14,6 @@ object Versions {
     const val JSON_WEBTOKEN = "0.11.5"
     const val JJWT_IMPL = "0.11.5"
     const val JJWT_JACKSON = "0.11.5"
+    const val CIRCUIT_BREAKER = "3.0.4"
+    const val ACTUATOR = "3.2.3"
```
adsukharev, line 18: Why do you need actuator ?

cloud/api-gateway/src/main/resources/application.yaml
Hoocki: no, it means all other settings work during the specified period (if 3 sent requests returned an error within 1 second, the service switches to OPEN mode)

buildSrc/src/main/kotlin/Versions.kt
Hoocki: to see the status of server and other settings in real time

adsukharev: Do you need it for circuit breaker ? Or just for yourself ?

Hoocki: just for me and i think i can move to local config

Hoocki:
I have no idea what the values should be, so I left the settings I tested with (except sliding-window-size)

register-health-indicator: provides information about the current status of the Circuit Breaker, such as the number of open/closed circuits, error percentage, etc.

sliding-window-type: defines type of sliding window for collecting statistics (TIME_BASED or COUNT_BASED)

sliding-window-size: size of the sliding window (based on type)

failure-rate-threshold: threshold value of the error percentage, at which the circuit enters the OPEN state.

minimum-number-of-calls: minimum number of calls that must be made before Circuit Breaker starts making decisions about switching between states. (from CLOSE to OPEN)

automatic-transition-from-open-to-half-open-enabled: determines automatic switching from the OPEN to the HALF-OPEN state is allowed.

wait-duration-in-open-state: waiting time in the OPEN state before switching to the HALF-OPEN state.

allowed-number-of-calls-in-half-open-state: maximum number of allowed calls in the HALF-OPEN state before returning to the CLOSE state.

adsukharev (approved):
In general numbers are not so important, since this project is not for prod

PR #66, opened 2024-03-21
[TSK-250] register services

Hoocki:


auth-manager/src/main/resources/application-local.yaml
```diff
@@ -9,6 +9,11 @@ server:
   error:
     include-message: always
 
+eureka:
+  client:
+    service-url:
+      defaultZone: http://localhost:8088/eureka
```
adsukharev: I would use default port for service discovery 8761

notification-manager/build.gradle.kts
```diff
@@ -4,20 +4,17 @@ plugins {
...
 	implementation("org.springframework.boot:spring-boot-starter-validation")
 	implementation("org.springframework.boot:spring-boot-starter-mail")
 	implementation("org.springframework.kafka:spring-kafka:${Versions.KAFKA}")
+	implementation("org.springframework.cloud:spring-cloud-starter-netflix-eureka-client")
+}
+
+dependencyManagement {
+	imports {
+		mavenBom("org.springframework.cloud:spring-cloud-dependencies:${Versions.SPRING_CLOUD}")
+	}
```
adsukharev, line 17: since you use everywhere spring boot web and spring cloud (or maybe some other spring libs), it makes sense to create a new plugin `spring.base` which includes `java.base` and these common dependencies.
For the libs module you don't need to use a new plugin, but others do

adsukharev (commented):
.

adsukharev (approved):

adsukharev (commented):
Don't forget to create a ticket to improve spring base plugin by moving actual spring boot plugins to your custom one

PR #69, opened 2024-03-24
[TSK-163] service discovery

Hoocki:


user-manager/build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 plugins {
     id("org.springframework.boot")
     id("io.spring.dependency-management")
```
adsukharev: We don't need these spring plugins anymore here and in all other modules where you use your custom one

adsukharev (approved):

kostya4201337 (approved):

PR #72, opened 2024-03-24
[TSK-243] add toggle for sending email

Hoocki:


notification-manager/src/main/java/com/qpay/notificationmanager/service/impl/MockNotificationService.java
```diff
@@ -0,0 +1,20 @@
...
+import com.qpay.libs.models.PaymentNotification;
+import com.qpay.notificationmanager.service.NotificationService;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Service;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
```
adsukharev: you don't need it

notification-manager/src/main/java/com/qpay/notificationmanager/service/impl/MockNotificationService.java
```diff
@@ -0,0 +1,20 @@
...
+@ConditionalOnProperty(name = "email.enabled", havingValue = "false")
+public class MockNotificationService implements NotificationService {
+
+    public void sendMessage(final PaymentNotification paymentNotification) {
+        log.info("Mock sending message to email: {}", paymentNotification.emailTo());
```
adsukharev:
```suggestion
        log.info("Mock email message sent to email: {}", paymentNotification.emailTo());
```

adsukharev (approved):

notification-manager/src/main/java/com/qpay/notificationmanager/service/impl/MockNotificationService.java
Hoocki: but without it i couldn't initialize service

adsukharev: I mean, you could use no args constructor to show you don't have any fields. but you can leave this one

Hoocki: ok

PR #78, opened 2024-03-28
[TSK-257] add top up support for transaction history

kostya4201337:


transaction-history-manager/src/main/java/com/qpay/transactionhistorymanager/model/TransactionType.java
```diff
@@ -0,0 +1,6 @@
+package com.qpay.transactionhistorymanager.model;
+
+public enum TransactionType {
+    TRANSFER,
```
adsukharev: For me, p2p (person to person) is a transfer, but p2b (person to business) is a payment.

adsukharev (approved):
@Hoocki all next PRs from Kostya about transactions will be reviewed firstly by you. After you approve it, I will check

Hoocki (approved):

PR #85, opened 2024-04-21
[TSK-267] add footer

Hoocki:


frontend/src/components/footer/Footer.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import {Container, Typography} from "@mui/material";
+import './styles.css';
+import {FOOTER_CONTENT} from "../../common/constansts/footer";
+
+const Footer = () => {
```
adsukharev: add FC type

frontend/src/common/constansts/footer.ts
```diff
@@ -0,0 +1 @@
+export const FOOTER_CONTENT = `© ${new Date().getFullYear()} Qpay. All rights are reserved`;
```
adsukharev:
```suggestion
export const FOOTER_CONTENT = `© ${new Date().getFullYear()} ${Titles.APP_NAME}. All rights are reserved`;
```

P.S I think Titles.APP_NAME should be "QPay" not "Qpay"

adsukharev (approved):

PR #86, opened 2024-04-22
[TSK-155] setup frontend structure

Hoocki:


frontend/src/App.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Providers from "./providers/Providers";
+import {StyledEngineProvider} from "@mui/material";
+
+const App: React.FC = () => {
+
+    return (
+        <StyledEngineProvider injectFirst>
```
adsukharev: This provider component should be in providers

frontend/src/components/header/mobileLogo/MobileLogo.tsx
```diff
@@ -0,0 +1,19 @@
...
+import {Typography} from "@mui/material";
+import './styles.css';
+import {Titles} from "../../../common/constansts/titles";
+
+const MobileLogo: React.FC = () => {
+    return (
+        <Typography
+            variant="h5"
+            noWrap
+            component="a"
```
adsukharev: Is "a" really what you want ? I'm just asking, I don't know

frontend/src/components/header/mobileMenu/MobileMenu.tsx
```diff
@@ -0,0 +1,47 @@
...
+                transformOrigin={{
+                    vertical: 'top',
+                    horizontal: 'left',
+                }}
+                open={Boolean(anchorEl)}
+                onClose={handleCloseMenu}
+                className="settings-menu"
+            >
+                {HEADER_MAIN_TABS.map((page) => (
+                    <MenuItem key={page} onClick={(event) => handleCloseMenu(event)}>
```
adsukharev:
```suggestion
                    <MenuItem key={page} onClick={handleCloseMenu}>
```
Can you try to do like this?
If it works, this approach is better as you don't need to create additional function, you just pass reference to existing one.
Check other places, where you can a little bit improve performance by using this approach

frontend/src/components/header/userMenu/UserMenu.tsx
```diff
@@ -0,0 +1,55 @@
...
+                    <Avatar/>
+                </IconButton>
+            </Tooltip>
+            <Menu
+                className="user-settings-menu"
+                id="menu-appbar"
+                anchorEl={anchorElUser}
+                anchorOrigin={{
+                    vertical: 'top',
+                    horizontal: 'right',
```
adsukharev: You use the same Params here and below. If you can extract it and reuse,It would be better

adsukharev (approved):

adsukharev:
Push Kostya to review faster. In the meantime, don't wait and work on the next tickets

frontend/src/components/header/mobileLogo/MobileLogo.tsx
Hoocki: no, it isn't necessary, forgot to delete it

adsukharev:
> Push Kostya to review faster. In the meantime, don't wait and work on the next tickets

you can work on the next frontend features. Just create new branch from this one (tsk 155)

kostya4201337 (approved):

PR #87, opened 2024-04-22
[TSK-269] add userId and userType in token

Hoocki:


auth-manager/src/main/java/com/qpay/authmanager/service/jwt/JwtService.java
```diff
@@ -1,8 +1,10 @@
 package com.qpay.authmanager.service.jwt;
 
+import com.qpay.libs.models.UserType;
+
 public interface JwtService {
 
-    String generateToken(String email);
+    String generateToken(String email, long userId, UserType userType);
```
adsukharev: It's better to create a new model, let's say TokenData and pass it here. It's more extendable. If you decide to add one more parameter later you don't need to change signature everywhere (interface, implementation, tests), you only add new field to this model.

auth-manager/src/main/java/com/qpay/authmanager/service/jwt/impl/JwtServiceImpl.java
```diff
@@ -21,14 +23,17 @@ public class JwtServiceImpl implements JwtService {
...
+        claims.put("userId", userId);
+        claims.put("userType", userType);
+        return createToken(claims);
     }
 
-    private String createToken(final Map<String, Object> claims, final String email) {
+    private String createToken(final Map<String, Object> claims) {
         return Jwts.builder()
                 .setClaims(claims)
-                .setSubject(email)
```
adsukharev, line 31: are you sure you don't need it ?
I have no idea

adsukharev (approved):

auth-manager/src/main/java/com/qpay/authmanager/service/jwt/impl/JwtServiceImpl.java
Hoocki: i put it in `claims` and then checked in JWT Io how it looks and everithing normal, except that now this field `email` not `sub`

kostya4201337 (approved):

PR #88, opened 2024-04-23
[TSK-276] add layout for sign up page

Hoocki:


frontend/src/common/constansts/messages.ts
```diff
@@ -0,0 +1,4 @@
+export enum Messages {
```
adsukharev: Message is the same as notifications. Find another name.

frontend/src/common/constansts/messages.ts
```diff
@@ -0,0 +1,4 @@
...
+    TERMS = "I agree with terms, conditions, and privacy policy",
+    ACCOUNT = "Already have an account? "
```
adsukharev:
```suggestion
    ACCOUNT_EXISTS = "Already have an account? "
```

adsukharev: these 2 variables are different . Move it to different enums

frontend/src/pages/auth/signin/SignIn.tsx
```diff
@@ -49,6 +50,18 @@ const SignIn: React.FC = () => {
...
+            <Typography
+                variant="body2"
+                sx={{ marginTop: 1 }}
+            >
+                Don't have an account?{' '}
+                <Link
+                    href={Paths.SignUp}
+                    color="primary"
+                >
+                    Sign up
```
adsukharev:
```suggestion
                   Buttons.SIGN_UP
```

frontend/src/pages/auth/signin/SignIn.tsx
```diff
@@ -49,6 +50,18 @@ const SignIn: React.FC = () => {
             >
                 {Buttons.SIGN_IN}
             </Button>
+            <Typography
+                variant="body2"
+                sx={{ marginTop: 1 }}
+            >
+                Don't have an account?{' '}
```
adsukharev: move to the same place, where you define `Already have an account?`

frontend/src/components/header/Header.tsx
```diff
@@ -46,13 +47,17 @@ const Header: React.FC<HeaderProps> = ({isLogged}: HeaderProps) => {
...
 
     const redirectToPage = (isLogged: boolean, navigate: NavigateFunction) => {
-        const path = isLogged ? Paths.Home : Paths.SignIn;
-        navigate(path);
+        if (!isLogged && location.pathname !== Paths.SignUp) {
+            navigate(Paths.SignIn);
+        }
+        else if (isLogged) {
+            navigate(Paths.Home);
+        }
```
adsukharev, line 54: if you check `if (isLogged)` on the first place and it's false, you don't need to check this variable again in `else if`.
```typescript
if (isLogged) {
            navigate(Paths.Home);
        } else if (location.pathname !== Paths.SignUp) {
            navigate(Paths.SignIn);
        }      
```

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+            setMerchantChecked(false);
+        }
+    };
+
+    const handleMerchantCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setMerchantChecked(event.target.checked);
+        if (event.target.checked) {
+            setCustomerChecked(false);
+        }
+    };
```
adsukharev, line 61: Checkbox is used when we want to check several options. However, if we allow a user to choose only one option, it's more convenient to use radio button for clients and devs because you don't need to manually handle each case: https://mui.com/material-ui/react-radio-button/

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+import {UserCredentials} from "../../../types/userCredentials";
+import {Buttons} from "../../../common/constansts/buttons";
+import {Messages} from "../../../common/constansts/messages";
+import {Roles} from "../../../common/constansts/roles";
+import {UserType} from "../../../types/userType";
 
 const SignUp: React.FC = () => {
+    const [email, setEmail] = useState('');
+    const [password, setPassword] = useState('');
+    const [name, setName] = useState('');
```
adsukharev: instead of creating several states, create common one which stores UserCredentials. Above the component you can create an initial state with empty fields.
When you need to set new value, e.g email, approach is same as in redux. You have to create new object of userCredentials  (see spread operator https://www.sitepoint.com/merging-objects-javascript/) and then replace target value, in our example `email`. For readability, we can use destructing and rename variable. https://wesbos.com/destructuring-renaming.

option 1.
```typescript
onChange={({target: {value: email}}) => setUserCredentials({...userCredentials, email})}
``` 

option 2.
```typescript
 onChange={(event) => setUserCredentials({...userCredentials, email: event.target.value})}
``` 

Both options are equal, but 1st one is more readable to my mind

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+    const handleAgreeCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setAgreeChecked(event.target.checked);
+    };
+
+    const handleSignUp = () => {
+        const isSignUpEnabled = email && name && password && (customerChecked || merchantChecked) && agreeChecked;
+        if (isSignUpEnabled) {
+            const userCredentials = createUserCredentials(email, name, password);
+            const userType = getUserType(customerChecked);
+            console.log("userCredentials: " + userCredentials + ", userType: " + userType);
```
adsukharev: remove

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+
+    const handleMerchantCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setMerchantChecked(event.target.checked);
+        if (event.target.checked) {
+            setCustomerChecked(false);
+        }
+    };
+
+    const handleAgreeCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setAgreeChecked(event.target.checked);
```
adsukharev: since it's only one line without any logic, you can create lambda function in place

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+    };
+
+    const handleSignUp = () => {
+        const isSignUpEnabled = email && name && password && (customerChecked || merchantChecked) && agreeChecked;
```
adsukharev: Create separate method which validates all cases.

Btw, your email and password are not validated correctly. Set some simple rules for password, e.g. at least 4 symbols. If you already have some rules on the backend, sync it. Same for email.

If you think, why do we need to validate it on the client if we do the same on the server the answer is simple, we don't want to bother/load our servers with unnecessary calls.

But if we validate these fields on the client why do we need to validate it on the server? We can never rely on what client sends us, because user can be hacked. Or we expose our api for 3rd party clients and they can do poor validation.

adsukharev:
```suggestion
        const isValid = email && name && password && (customerChecked || merchantChecked) && agreeChecked;
```

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+        if (isSignUpEnabled) {
+            const userCredentials = createUserCredentials(email, name, password);
+            const userType = getUserType(customerChecked);
+            console.log("userCredentials: " + userCredentials + ", userType: " + userType);
+            navigate(Paths.SignIn);
```
adsukharev: add mock call before navigating

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
-            {Titles.SIGN_UP}
-        </p>
+        <Box className="box">
+            <Box className="box-inner">
+                <Box
+                    component="img"
+                    className="logo"
+                    alt={Titles.APP_NAME}
+                    src={Logo}
+                />
```
adsukharev, line 89: you are going to use the same icon component for sign in, so it's better to extract it

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+
+                <TextField
+                    placeholder="Email"
```
adsukharev: provide feedback to the user about the error. `error={isEmailValid}` 
 https://mui.com/material-ui/react-text-field/#validation

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+                    type="password"
+                    variant="outlined"
+                    value={password}
+                    onChange={(e) => setPassword(e.target.value)}
+                    fullWidth
+                    margin="normal"
+                />
+
+                <Typography variant="h5" gutterBottom>
+                    Select role
```
adsukharev: constant

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,153 @@
...
+                />
+
+                <TextField
+                    placeholder="Email"
+                    variant="outlined"
+                    value={email}
+                    onChange={(e) => setEmail(e.target.value)}
+                    fullWidth
+                    margin="normal"
+                />
```
adsukharev, line 93: You are going to use email, password at least 3 times in your app, name 2 times. Let's extract these common components. 

Each component should store it's value state and validation result. Default value should be passed from parent component and be set only once in the beginning. https://react.dev/reference/react/useEffect#examples-dependencies.
I think you can use smth like this 
```typescript
useEffect(() => {
        setEmail(defaultEmail)
    }, []); // empty array means you run this hook only once during first rendering of this component
``` 

Then pass both values to the parent component in your case SignUp using callback during typing. When you receive value (e.g. email or password) and validation result ( e.g. isEmailValid), you update UserCredentials and isValid accordingly in parent component.

frontend/src/pages/auth/signup/styles.css
```diff
@@ -0,0 +1,46 @@
+.box {
```
adsukharev: 1) find better name, for other styles as well.
2) some of the styles are going to be reused in SingIn component. Move it to auth folder

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,147 +1,136 @@
...
-    const handleSignUp = () => {
-        const isSignUpEnabled = email && name && password && (customerChecked || merchantChecked) && agreeChecked;
-        if (isSignUpEnabled) {
-            const userCredentials = createUserCredentials(email, name, password);
-            const userType = getUserType(customerChecked);
-            console.log("userCredentials: " + userCredentials + ", userType: " + userType);
+    const handleSignUp = async () => {
+        const isValid = validateData();
+        if (isValid) {
+            await addUserService(userCredentials, selectedRole);
```
adsukharev: it's good to tell a user that he registered successfully

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,147 +1,136 @@
...
-                    fullWidth
-                    margin="normal"
+    return (
+        <Box className="auth-container">
+            <Box className="auth-content">
+                <AuthLogo/>
+                <EmailUserCredentials
+                    defaultEmail={userCredentials.email}
+                    updateEmail={updateEmailChange}
+                    defaultEmailError={emailError}
```
adsukharev: it does not make sense to pass this parameter

frontend/src/common/constansts/errorMessages.ts
```diff
@@ -0,0 +1,5 @@
+export enum ErrorMessages {
```
adsukharev:
```suggestion
export enum ValidationErrorMessages {
```

frontend/src/common/constansts/errorMessages.ts
```diff
@@ -0,0 +1,5 @@
...
+    EMAIL_NOT_VALID = "Please enter a valid email address",
+    NAME_NOT_VALID = "Name has to contain at least 4 symbols",
```
adsukharev:
```suggestion
    NAME_NOT_VALID = "Name must contain at least 4 symbols",
```

frontend/src/services/user.ts
```diff
@@ -18,4 +19,23 @@ export const getUserService = async (userId: number, userType: UserType): Promis
         createdAt: new Date(),
         updatedAt: new Date(),
     }
+}
+
+export const addUserService = async (userCredentials: UserCredentials, userType: UserType): Promise<User> => {
```
adsukharev:
```suggestion
export const signUpService = async (userCredentials: UserCredentials, userType: UserType): Promise<User> => {
```

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,142 @@
...
+                    <Typography variant="body2">
+                        {AuthContent.TERMS}
+                    </Typography>
+                </Box>
+
+                <Button
+                    onClick={handleSignUp}
+                    variant="contained"
+                    className="button"
+                    disabled={!selectedRole || !agreeChecked}
```
adsukharev: incorrect validation. You allow to click the button if name, email, password are empty. I think you did it on purpose, why?

frontend/src/pages/auth/components/email/EmailUserCredentials.tsx
adsukharev, line 1: components/fields/
then move there your custom email, name, password components

frontend/src/pages/auth/components/email/EmailUserCredentials.tsx
```diff
@@ -0,0 +1,43 @@
+import React, {useEffect, useState} from 'react';
+import {ErrorMessages} from "../../../../common/constansts/errorMessages";
+import {TextField} from "@mui/material";
+import {EmailUserCredentialsProps} from "./props";
+import {EMAIL_REGEX} from "../../../../common/constansts/emailRegex";
+
+const EmailUserCredentials: React.FC<EmailUserCredentialsProps> = ({defaultEmail, updateEmail, defaultEmailError, updateEmailError}: EmailUserCredentialsProps) => {
```
adsukharev:
```suggestion
const EmailField: React.FC<EmailUserCredentialsProps> = ({defaultEmail, updateEmail, defaultEmailError, updateEmailError}: EmailUserCredentialsProps) => {
```

frontend/src/pages/auth/components/email/EmailUserCredentials.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    const [email, setEmail] = useState<string>(defaultEmail);
+    const [emailError, setEmailError] = useState<boolean>(defaultEmailError);
```
adsukharev:
```suggestion
    const [isEmailValid, setIsEmailValid] = useState<boolean>(false);
```

frontend/src/pages/auth/components/email/EmailUserCredentials.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    const handleEmailChange = (email: string) => {
+        setEmail(email);
+        validateEmail(email);
+    };
+
+    const validateEmail = (email: string) => {
```
adsukharev: your validate method must only validate data. But you do lots of things, I assume this method potentially can bring a cup of tee

frontend/src/pages/auth/components/email/EmailUserCredentials.tsx
```diff
@@ -0,0 +1,43 @@
...
+        const isValid = email.trim().match(EMAIL_REGEX) !== null;
+        setEmailError(!isValid);
+        updateEmail(email);
+        updateEmailError(isValid);
+    };
+
+    useEffect(() => {
+        setEmail(defaultEmail);
+        setEmailError(defaultEmailError);
+    }, []);
```
adsukharev: try this code.

```typescript
    const [email, setEmail] = useState<string>(defaultEmail);
    const [isValid, setIsValid] = useState<boolean>(false);

    const handleEmailChange = (email: string) => {
        setEmail(email);
        const isEmailValid = validateEmail(email);
        setIsValid(isEmailValid);
        updateEmailFields({email, isEmailValid}); // of course we can pass isValid, but state setters work async. It means there's no guarantee that we pass fresh data. It's safer to pass local variable isEmailValid
    };

    const validateEmail = (email: string): boolean => email.trim().match(EMAIL_REGEX) !== null;

    useEffect(() => {
        setEmail(defaultEmail);
        validateEmail(defaultEmail);
    }, []);

    return (
        <TextField
            error={!isValid}
            helperText={isValid ? "" :  ErrorMessages.EMAIL_NOT_VALID}
            value={email}
            placeholder="Email"
            variant="outlined"
            onChange={(e) => handleEmailChange(e.target.value)}
            fullWidth
            margin="normal"
        />
    );
```

frontend/src/pages/auth/components/email/props.ts
```diff
@@ -0,0 +1,6 @@
+export interface EmailUserCredentialsProps {
+    defaultEmail: string;
+    updateEmail: (email: string) => void;
+    defaultEmailError: boolean;
+    updateEmailError: (emailError: boolean) => void;
```
adsukharev: updateEmailError and updateEmail are triggered at the same time for each type. You can merge them and pass both params at once

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,142 @@
...
+    password: ""
+}
 
 const SignUp: React.FC = () => {
+    const [userCredentials, setUserCredentials] = useState<UserCredentials>(initialUserCredentials);
+    const [selectedRole, setSelectedRole] = useState<UserType>(UserType.Customer);
+    const [agreeChecked, setAgreeChecked] = useState(false);
+    const [emailError, setEmailError] = useState(false);
+    const [nameError, setNameError] = useState(false);
+    const [passwordError, setPasswordError] = useState(false);
```
adsukharev: 1) 
create common state for validation results:
```typescript
const [fieldsValidation, ...] = 
{
 "isEmailValid": false,
...
}
``` 
you can create types.ts at local folder since this type is only for SignUp component usage
2) create isValid state.
3) each time when fieldsValidation is changed or agreeChecked or selectedRole, update isValid

frontend/src/types/userCredentials.ts
adsukharev, line 1: merge all users types (userType, userCredentials, loggedUser -> user.ts). Same for notifications

frontend/src/pages/auth/signin/SignIn.tsx
adsukharev, line 1: `const notification = createNotification(Notifications.SIGN_UP, NotificationType.SUCCESS);`

I think we don't need to notify a user about successful login. If they redirected to main page, it's clear that they logged in

frontend/src/pages/auth/components/fields/password/props.ts
```diff
@@ -0,0 +1,4 @@
+export interface PasswordUserCredentialsProps {
+    defaultPassword: string;
```
adsukharev: we never return password from server, so it's not possible to provide default password. User has to type it each time from the scratch

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,156 @@
...
+    const dispatch = useAppDispatch();
+
+    const handleRoleChange = (event: React.ChangeEvent<HTMLInputElement>) => setSelectedRole(event.target.value as UserType);
+    const handleAgreeCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => setAgreeChecked(event.target.checked);
+
+    const handleSignUp = async () => {
+        if (isValid) {
+            await signUpService(userCredentials, selectedRole);
+            navigate(Paths.SignIn);
+            const notification = createNotification(Notifications.SIGN_UP, NotificationType.SUCCESS);
```
adsukharev, line 51: send notification, then route to the next page

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,156 @@
...
+            dispatch(showNotification(notification));
+        } else {
+            const notification = createNotification(Notifications.VALIDATION_NOT_PASSED, NotificationType.ERROR);
```
adsukharev: are you sure that the only error is VALIDATION_NOT_PASSED during registration? Maybe email already exists or server is down

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,156 @@
...
+            dispatch(showNotification(notification));
+        }
+    };
+
+    const updateEmailChange = (email: string, isEmailValid: boolean) => {
```
adsukharev, line 58: here you name variable as isEmailValid, but in interface just isValid. Sync it

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,156 @@
...
+    };
+
+    useEffect(() => {
+        setIsValid(
+            fieldsValidation.isEmailValid &&
+            fieldsValidation.isPasswordValid &&
+            fieldsValidation.isNameValid &&
+            selectedRole &&
+            agreeChecked
+        );
```
adsukharev: useEffect should not have logic. The only task is to call relevant methods when some deps are changed . Otherwise during the time it can be very big and hard to read

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,156 @@
...
+    const updatePasswordChange = (password: string, isPasswordValid: boolean) => {
+        setUserCredentials({...userCredentials, password});
+        setFieldsValidation({...fieldsValidation, isPasswordValid});
+    };
+
+    useEffect(() => {
+        setIsValid(
+            fieldsValidation.isEmailValid &&
+            fieldsValidation.isPasswordValid &&
+            fieldsValidation.isNameValid &&
```
adsukharev: come up with a better approach. Try to check that all fields of fieldsValidation object are valid without manual selection of fields. If later you add one more field, you don't need to update the code.

frontend/src/common/constansts/textFieldValidation.ts
adsukharev, line 1: I would rename file to just validation

frontend/src/common/constansts/emailRegex.ts
```diff
@@ -0,0 +1 @@
+export const EMAIL_REGEX = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|.(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
adsukharev: move it to validation constant

frontend/src/pages/auth/components/fields/email/props.ts
```diff
@@ -0,0 +1,4 @@
+export interface EmailUserCredentialsProps {
```
adsukharev:
```suggestion
export interface EmailFieldProps {
```
same for others

frontend/src/pages/auth/components/fields/name/NameField.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    useEffect(() => {
+        setName(defaultName);
+        validateName(defaultName);
+    }, []);
+
+    return (
+        <TextField
+            error={!isNameValid}
+            helperText={!isNameValid ? ValidationErrorMessages.NAME_NOT_VALID : ""}
```
adsukharev:
```suggestion
            helperText={isNameValid ? "" : ValidationErrorMessages.NAME_NOT_VALID}
```

If we can reduce amount of inversions, conditions -> it's easier to read

frontend/src/pages/auth/components/fields/password/PasswordField.tsx
adsukharev, line 1: You are going to use it for profile as well. Move it to `/components/` folder. Not `/pages/auth/components`

frontend/src/pages/auth/components/fields/password/PasswordField.tsx
```diff
@@ -0,0 +1,43 @@
+import React, {useEffect, useState} from 'react';
+import {ValidationErrorMessages} from "../../../../../common/constansts/validationErrorMessages";
+import {TextField} from "@mui/material";
+import {PasswordUserCredentialsProps} from "./props";
+import {TEXT_FIELD_VALIDATION} from "../../../../../common/constansts/textFieldValidation";
+
+const PasswordField: React.FC<PasswordUserCredentialsProps> = ({
+                                                                   defaultPassword,
+                                                                   updatePasswordFields
+                                                               }: PasswordUserCredentialsProps) => {
```
adsukharev:
```suggestion
                                                               }) => {
```

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,156 @@
...
+                        />
+                        <Typography variant="body2">{Roles.CUSTOMER}</Typography>
+                    </Box>
+                    <Box className="role-option">
+                        <Radio
+                            checked={selectedRole === UserType.Merchant}
+                            onChange={handleRoleChange}
+                            name="radio-buttons"
+                            value={UserType.Merchant}
+                        />
```
adsukharev: Why did you decide to use manual approach of creating separate radio buttons instead of automatic radio group ?
https://mui.com/material-ui/react-radio-button/#direction

adsukharev (commented):
Don't forget to check time to time how it looks like on mobile phone

frontend/src/pages/auth/signup/SignUp.tsx
Hoocki: thought, it would be easier to change component from `Checkbox` to `Radio`, but i will try to use automatic approach

frontend/src/components/password/PasswordField.tsx
```diff
@@ -0,0 +1,44 @@
...
+import {TextField} from "@mui/material";
+import {PasswordFieldProps} from "./props";
+import {TEXT_FIELD_VALIDATION} from "../../common/constansts/validation";
+
+const initialPassword = "";
+
+const PasswordField: React.FC<PasswordFieldProps> = ({
+                                                         updatePasswordFields
+                                                     }) => {
+    const [password, setPassword] = useState<string>(initialPassword);
```
adsukharev:
```suggestion
    const [password, setPassword] = useState<string>("");
```

frontend/src/components/password/PasswordField.tsx
```diff
@@ -0,0 +1,44 @@
...
+        setIsPasswordValid(isValid);
+        updatePasswordFields(password, isValid);
+    };
+
+    const validatePassword = (password: string) => password.trim().length >= TEXT_FIELD_VALIDATION;
+
+    useEffect(() => {
+        setPassword(initialPassword);
+        validatePassword(initialPassword);
+    }, []);
```
adsukharev: there's nothing to set and validate. Remove use effect

frontend/src/pages/auth/components/fields/email/EmailField.tsx
```diff
@@ -0,0 +1,43 @@
+import React, {useEffect, useState} from 'react';
+import {ValidationErrorMessages} from "../../../../../common/constansts/validationErrorMessages";
+import {TextField} from "@mui/material";
+import {EmailFieldProps} from "./props";
+import {EMAIL_REGEX} from "../../../../../common/constansts/validation";
+
+const EmailField: React.FC<EmailFieldProps> = ({
+                                                   defaultEmail,
+                                                   updateEmailFields
+                                               }) => {
```
adsukharev: I would do it inline, as there's enough space 
`const EmailField: React.FC<EmailUserCredentialsProps> = ({defaultEmail, updateEmailFields}) => {`

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,144 @@
...
+        setUserCredentials({...userCredentials, name});
+        setFieldsValidation({...fieldsValidation, isNameValid});
+    };
+
+    const updatePasswordChange = (password: string, isPasswordValid: boolean) => {
+        setUserCredentials({...userCredentials, password});
+        setFieldsValidation({...fieldsValidation, isPasswordValid});
+    };
+
+    const validateDate = () => {
```
adsukharev:
```suggestion
    const validateData = () => {
```

frontend/src/pages/auth/signup/SignUp.tsx
```diff
@@ -1,12 +1,144 @@
...
+        const isAllFieldsValid = Object.values(fieldsValidation).every(field => field);
```
adsukharev:
```suggestion
        const areAllFieldsValid = Object.values(fieldsValidation).every(field => field);
```
or
```suggestion
        const fieldsValid = Object.values(fieldsValidation).every(field => field);
```

adsukharev (approved):

PR #90, opened 2024-04-27
[TSK-270] add connection with the server

Hoocki:
add subject in token because it doesn't want to validate (but i checked previously that `subject` isn't required, so i'm at loss)


frontend/src/common/utils.ts
```diff
@@ -1,10 +1,6 @@
...
-import {UserType} from "../types/user";
+import {jwtDecode} from "jwt-decode";
 
 export const decodeToken = (token: string): TokenData => {
-    return {
-        userId: 0,
-        email: "user@gmail.com",
-        userType: UserType.Customer
-    };
+    return jwtDecode(token);
```
adsukharev: no need to have decodeToken method if you have inside another inline method

frontend/src/common/constansts/paths.ts
```diff
@@ -5,4 +5,12 @@ export enum Paths {
     Home = "/home",
     SignIn = "/signin",
     SignUp = "/signup"
+}
+
+export enum HttpPaths {
```
adsukharev: I would separate paths for different needs. Auth, users, payment and so on

frontend/src/stores/redux/loggedUser/loggedUserSlice.ts
```diff
@@ -15,11 +15,14 @@ export const loggedUserSlice = createSlice({
     initialState,
     reducers: {
         signIn: (state, action: PayloadAction<ILoggedUser>) => ({...action.payload}),
+        addToken: (state, action: PayloadAction<string>) => {
+            state.token = action.payload;
```
adsukharev, line 19: where do you use it?

frontend/src/services/user.ts
```diff
@@ -1,39 +1,22 @@
...
-            updatedAt: new Date(),
-        }
-    }
-    return {
-        id: 1,
-        email: "usermerchant@gmail.com",
-        name: "UserMerchant",
-        createdAt: new Date(),
-        updatedAt: new Date(),
+        const response = await httpClient.post<User>(HttpPaths.CUSTOMER + HttpPaths.SIGN_UP_USER, userCredentials);
```
adsukharev: why do you use here string + string, but above `${string}/${stirng}` ?

frontend/src/services/user.ts
```diff
@@ -1,39 +1,22 @@
...
-        id: userId,
-        email: "usermerchant@gmail.com",
-        name: "UserMerchant",
-        createdAt: new Date(),
-        updatedAt: new Date(),
+        const response = await httpClient.get<User>(`${HttpPaths.CUSTOMER}/${userId}`);
+        return _.get(response, "data");
     }
+    const response = await httpClient.get<User>(`${HttpPaths.MERCHANT}/${userId}`);
+    return _.get(response, "data");
```
adsukharev, line 10: you can extract "data" to a constant or move this line to utils file and reuse it everywhere.

auth-manager/src/main/java/com/qpay/authmanager/service/jwt/impl/JwtServiceImpl.java
```diff
@@ -34,6 +34,7 @@ public String generateToken(final TokenData tokenData) {
     private String createToken(final Map<String, Object> claims) {
         return Jwts.builder()
                 .setClaims(claims)
+                .setSubject(claims.get("email").toString())
```
adsukharev, line 37: why did use decide move it back?

frontend/src/services/user.ts
```diff
@@ -1,39 +1,22 @@
 import {User, UserCredentials, UserType} from "../types/user";
+import httpClient from "./httpClient";
+import {HttpPaths} from "../common/constansts/paths";
+import _ from "lodash";
 
 export const getUserService = async (userId: number, userType: UserType): Promise<User> => {
     if (userType === UserType.Customer) {
```
adsukharev:
```typescript
const basePath = userType === UserType.Customer ? HttpPaths.CUSTOMER : HttpPaths.MERCHANT;
const targetPath = `${basePath}/${userId}`;
const response = await httpClient.get<User>(targetPath);
return _.get(response, "data");
```

auth-manager/src/main/java/com/qpay/authmanager/service/jwt/impl/JwtServiceImpl.java
Hoocki: https://github.com/Hoocki/qpay/pull/90#issue-2266958905

frontend/src/stores/redux/loggedUser/loggedUserSlice.ts
Hoocki: if i want to create `loggedUser `i need to get `user`, but if i want to get user i need to provide token, so i add this reducer for this

frontend/src/services/user.ts
Hoocki: in get method i provide only request param(userId) without body, in post method different situation, use body without any param in url

frontend/src/stores/redux/loggedUser/loggedUserSlice.ts
adsukharev: not clear what you said

Hoocki: forgot to add to call this reducer in code (idk why i deleted it), so i will explain more detailed: 
1) when i get token i need to save it in store to get it in httpClient
2) when i call getUserService i used interceptor that add token in header from store
3) when i get user from server, i create loggedUser which will be contained in store

Hoocki: correct version of fucntion:

const handleSignIn = async () => {
        if (!isValid) return;
        const token = await logInService(authCredentials);
        if (token === "") return;
        dispatch(addToken(token));
        const decodedToken = decodeToken(token);
        const user = await getUserService(decodedToken.userId, decodedToken.userType);
        const loggedUser = createLoggedUser(token, user);
        if (loggedUser) {
            dispatch(signIn(loggedUser));
        }

    }

frontend/src/pages/auth/signin/SignIn.tsx
```diff
@@ -41,13 +42,14 @@ const SignIn: React.FC = () => {
     const handleSignIn = async () => {
         if (!isValid) return;
         const token = await logInService(authCredentials);
-        const decodedToken = decodeToken(token);
+        if (token === "") return;
```
adsukharev, line 45: Actually if token is empty , smth goes wrong and you have to throw an error. But you intercept errors, so I don't think you have to check it

frontend/src/services/user.ts
adsukharev: What about this comment ?

adsukharev (approved):

frontend/src/services/user.ts
Hoocki: in this case, idk how i get data from logInService(), because i extract data: "data.token" and for some cases i don't need to return default value or return default value based on type (null, empty string or smt else)

PR #91, opened 2024-04-28
[TSK-268] add confirmation before log out

Hoocki:


frontend/src/common/constansts/titles.ts
```diff
@@ -10,5 +10,6 @@ export enum Titles {
     TOP_UP = "Top up",
     LOG_OUT = "Log out",
     SETTINGS_HINT = "Open settings",
-    SELECT_ROLES = "Select role"
+    SELECT_ROLES = "Select role",
+    DIALOG = "Log out"
```
adsukharev: dialog is too general

frontend/src/components/header/userMenu/UserMenu.tsx
```diff
@@ -38,11 +73,30 @@ const UserMenu: React.FC<UserMenuProps> = ({anchorElUser, handleCloseMenu, handl
...
+                        {Notifications.CONFIRM_LOG_OUT}
+                    </DialogContentText>
+                </DialogContent>
+                <DialogActions>
+                    <Button variant="contained" color="error" onClick={handleClose}>{Buttons.Cancel}</Button>
+                    <Button variant="contained" onClick={handleLogOut} autoFocus>
+                        {Buttons.CONFIRM}
+                    </Button>
+                </DialogActions>
+            </Dialog>
```
adsukharev: move it to a dedicated component

adsukharev (commented):
1. Your signUpService method is in incorrect file
2. `const basePath = userType === UserType.Customer ? (UserPaths.CUSTOMER + UserPaths.SIGN_UP_USER) : (UserPaths.MERCHANT + UserPaths.SIGN_UP_USER);
` make it better. Check how you did it for `getUserService`

adsukharev:
add constant mockEnabled boolean.
Then in login and get user service based on this variable return either mock data or real request. It can make test process of your app easier. We don't need to run servers all the time. Same for all your future services.
Store mock data in services/mock.ts

frontend/src/components/header/userMenu/confirmLogOut/props.ts
```diff
@@ -0,0 +1,5 @@
+export interface ConfirmLogOutProps {
+    handleLogOut: () => void;
+    showDialog: boolean;
+    handleClose: () => void;
```
adsukharev, line 4: I think handleClose and showDialog can be controlled inside of ConfirmLogOut component. Only handleLogOut should be exposed

adsukharev (approved):

PR #95, opened 2024-05-01
[TSK-289] update routes

Hoocki:


frontend/src/routes/RouteAuth.tsx
```diff
@@ -1,13 +1,16 @@
...
 import {authRoutes, IRoute} from "./routesConfig";
+import {Paths} from "../common/constansts/paths";
 
 const RouteAuth: React.FC = () => {
     return (
         <Routes>
             {authRoutes.map((route: IRoute) => (
                 <Route key={route.path} path={route.path} element={route.element}/>
             ))}
+            <Route path="/*" element={<Navigate to={Paths.SignIn} replace={true} />}>
```
adsukharev:
```suggestion
            <Route path="/*" element={<Navigate to={Paths.SignIn} replace={true} />}/>
```

frontend/src/routes/RouteMain.tsx
```diff
@@ -1,13 +1,16 @@
...
 import {IRoute, mainRoutes} from "./routesConfig";
+import {Paths} from "../common/constansts/paths";
 
 const RouteMain: React.FC = () => {
     return (
         <Routes>
             {mainRoutes.map((route: IRoute) => (
                 <Route key={route.path} path={route.path} element={route.element}/>
             ))}
+            <Route path="/*" element={<Navigate to={Paths.Home} replace={true}/>}>
```
adsukharev:
```suggestion
            <Route path="/*" element={<Navigate to={Paths.Home} replace={true}/>}/>
```

adsukharev: add constant with Path.any

adsukharev (commented):
1) change in routes the way you render pages. If user logged in, you have to check type of the user. If client, render RoutesMain (change name to RoutesClientMain), if merchant, render RoutesMerchantMain which has only one Route to merchant home page. 
2) Don't show some links in header if user is merchant

frontend/src/routes/RouteMain.tsx
```diff
@@ -1,13 +1,22 @@
...
     return (
         <Routes>
-            {mainRoutes.map((route: IRoute) => (
+            {loggedUser.userType === UserType.Customer ? mainClientRoutes.map((route: IRoute) => (
                 <Route key={route.path} path={route.path} element={route.element}/>
-            ))}
+            )) : mainMerchantRoutes.map((route: IRoute) => (
+                <Route key={route.path} path={route.path} element={route.element}/>
+            ))
+            }
```
adsukharev: 1) it's impossible to read. 
2) you duplicate the code

Create smth like this:
```typescript
const getInitialState = (loggedUser: ILoggedUser) => loggedUser.userType === UserType.Customer ? mainClientRoutes : mainMerchantRoutes;

const RouteMain: React.FC = () => {
    const loggedUser = useAppSelector(selectLoggedUser); // instead of loggedUser get only userType. This method of getting only type you can move to the slice, as you are going to reuse it in other places
    const [mainRoutes] = useState<IRoute[]>(getInitialState(loggedUser));
    
    return (
        <Routes>
            {mainRoutes.map((route: IRoute) => 
                <Route key={route.path} path={route.path} element={route.element}/>)}
            <Route path={Paths.ANY} element={<Navigate to={Paths.Home} replace={true}/>}/>
        </Routes>
    );
}

export default RouteMain;
``` 

same for other cases across the app

frontend/src/routes/GlobalRoutes.tsx
```diff
@@ -1,25 +1,16 @@
...
-    const loggedUser = useAppSelector(selectLoggedUser);
-
-    const updateIsLogged = () => {
-        setIsLogged(!_.isEmpty(loggedUser.token));
-    }
-
-    useEffect(() => {
-        updateIsLogged();
-    }, [loggedUser]);
+    const isLogged = useAppSelector((state: RootState) => !_.isEmpty(state.loggedUserState.token));
```
adsukharev: Maybe it makes sense to move this method to slice to reuse it somewhere else in the future

frontend/src/components/header/Header.tsx
```diff
@@ -37,21 +35,9 @@ const Header: React.FC<HeaderProps> = ({isLogged}: HeaderProps) => {
 
     const findRouteByPath = (path: string) => {
         const normalizedPath = path.toLowerCase().replace(/\s/g, '');
-        return mainRoutes.find((route: IRoute) => route.path.trim().slice(1).toLowerCase() === normalizedPath);
+        return mainClientRoutes.find((route: IRoute) => route.path.trim().slice(1).toLowerCase() === normalizedPath);
```
adsukharev: I would make it more generic by selecting routes based on user type. Who knows, maybe merchant will have more tabs in header later

frontend/src/common/constansts/headers.ts
```diff
@@ -1,10 +1,18 @@
-export const HEADER_MAIN_TABS: string[] = [
+export const HEADER_CLIENT_MAIN_TABS: string[] = [
     "Home",
```
adsukharev: Create each tab as a separate constant, then use constant in arrays.

frontend/src/components/header/mobileMenu/MobileMenu.tsx
```diff
@@ -29,11 +37,16 @@ const MobileMenu: React.FC<MobileMenuProps> = ({ anchorEl, handleCloseMenu, hand
                 onClose={handleCloseMenu}
                 className="settings-menu"
             >
-                {HEADER_MAIN_TABS.map((page) => (
+                {loggedUser.userType === UserType.Customer ? HEADER_CLIENT_MAIN_TABS.map((page) => (
+                    <MenuItem key={page} onClick={handleCloseMenu}>
+                        <Typography className="typography">{page}</Typography>
+                    </MenuItem>
+                )) : HEADER_MERCHANT_MAIN_TABS.map((page) => (
```
adsukharev: It makes sense to pass tabs from header component to not duplicate logic here and in desctopMenu compoennt

frontend/src/common/constansts/paths.ts
adsukharev, line 1: split frontend paths and backend endpoints

adsukharev: and in general, if you have some common changes from previous PR which can be important now, it's better to make it here and merge it to reduce amount of conflicts in the future

frontend/src/common/constansts/headers.ts
```diff
@@ -1,10 +1,5 @@
-export const HEADER_MAIN_TABS: string[] = [
-    "Home",
-    "Top Up",
-    "Transactions",
-];
-
-export const HEADER_SETTINGS_TABS: string[] = [
-    "Profile",
-    "Log out"
-];
```
adsukharev, line 10: roll it back and instead of strings use constants

frontend/src/components/header/Header.tsx
```diff
@@ -1,22 +1,33 @@
...
 import DesktopMenu from "./desktopMenu/DesktopMenu";
 import UserMenu from "./userMenu/UserMenu";
 import './styles.css';
-import {Paths} from "../../common/constansts/paths";
+import {useAppSelector} from "../../stores/redux/hooks";
+import {RootState} from "../../stores/redux/store";
+import {UserType} from "../../types/user";
+import {HOME_TAB, LOG_OUT_TAB, PROFILE_TAB, TOP_UP_TAB, TRANSACTIONS_TAB} from "../../common/constansts/headers";
+
+const getInitialState = (userType: UserType) => userType === UserType.Customer ? mainClientRoutes : mainMerchantRoutes;
```
adsukharev: initial states of what?

frontend/src/components/header/Header.tsx
```diff
@@ -36,21 +47,16 @@ const Header: React.FC<HeaderProps> = ({isLogged}: HeaderProps) => {
...
-        }
-    };
-
     useEffect(() => {
-        redirectToPage(isLogged, navigate);
-    }, [isLogged]);
+        setMainTabs(getInitialStateMainTabs(loggedUserType));
+        setSettingsTabs(getInitialStateSettingsTabs(loggedUserType));
+        setMainRoutes(getInitialState(loggedUserType));
+    }, [loggedUserType]);
```
adsukharev, line 64: in what cases does user change his type?

frontend/src/components/header/Header.tsx
```diff
@@ -1,22 +1,33 @@
...
 
     const [anchorElNav, setAnchorElNav] = useState<null | HTMLElement>(null);
     const [anchorElUser, setAnchorElUser] = useState<null | HTMLElement>(null);
     const navigate = useNavigate();
-    const location = useLocation();
+    const loggedUserType = useAppSelector((state: RootState) => state.loggedUserState.userType);
+    const [mainRoutes, setMainRoutes] = useState<IRoute[]>(getInitialState(loggedUserType));
+    const [mainTabs, setMainTabs] = useState<string[]>(getInitialStateMainTabs(loggedUserType));
+    const [settingsTabs, setSettingsTabs] = useState<string[]>(getInitialStateSettingsTabs(loggedUserType));
+    console.log(loggedUserType);
```
adsukharev: 1. Review your code before pushing changes
2. push changes
3. review again on github

frontend/src/routes/GlobalRoutes.tsx
```diff
@@ -1,30 +1,19 @@
...
-    const loggedUser = useAppSelector(selectLoggedUser);
-
-    const updateIsLogged = () => {
-        setIsLogged(!_.isEmpty(loggedUser.token));
-    }
-
-    useEffect(() => {
-        updateIsLogged();
-    }, [loggedUser]);
+    const isUserLogged = useAppSelector(isLogged);
```
adsukharev: change name of the method `isLogged` it sounds like a variable but it is not. Moreover, it makes you change your real variables.

frontend/src/components/header/Header.tsx
```diff
@@ -1,22 +1,33 @@
...
+const getInitialStateMainTabs = (userType: UserType) => userType === UserType.Customer ? [HOME_TAB, TOP_UP_TAB, TRANSACTIONS_TAB] : [];
+const getInitialStateSettingsTabs = (userType: UserType) => userType === UserType.Customer ? [PROFILE_TAB, LOG_OUT_TAB] : [LOG_OUT_TAB];
 
 const Header: React.FC<HeaderProps> = ({isLogged}: HeaderProps) => {
 
     const [anchorElNav, setAnchorElNav] = useState<null | HTMLElement>(null);
     const [anchorElUser, setAnchorElUser] = useState<null | HTMLElement>(null);
     const navigate = useNavigate();
-    const location = useLocation();
+    const loggedUserType = useAppSelector((state: RootState) => state.loggedUserState.userType);
```
adsukharev: It's worth to extract this lambda to slice

frontend/src/components/header/Header.tsx
Hoocki: when i logged out and change userType, it doesn't refresh data, but if i manually reload it's updated

frontend/src/common/constansts/headers.ts
adsukharev: If you don't understand, ask.  You are adding and removing again and again. 
You have to instead of strings use constants. 
```typescript
export const LOG_OUT_TAB = "Log out";

export const HEADER_MERCHANT_SETTINGS_TABS: string[] = [
    LOG_OUT_TAB
];

export const HEADER_CLIENT_SETTINGS_TABS: string[] = [
    PROFILE_TAB,
    LOG_OUT_TAB
];

```

frontend/src/routes/GlobalRoutes.tsx
```diff
@@ -1,25 +1,14 @@
...
-    const loggedUser = useAppSelector(selectLoggedUser);
-
-    const updateIsLogged = () => {
-        setIsLogged(!_.isEmpty(loggedUser.token));
-    }
-
-    useEffect(() => {
-        updateIsLogged();
-    }, [loggedUser]);
+    const isLogged = useAppSelector(isUserLogged);
```
adsukharev: instead of renaming you just changed names.
isUserLogged is not what method does. It's name of the variable, but you have a method. You can name method same as you have in the example: `selectIsLogged`

adsukharev (approved):

PR #98, opened 2024-05-06
[TSK-264] refactor

adsukharev:
@Hoocki check the pattern of creating branches (again) TSK-ID_feature-name


frontend/src/common/constansts/time.ts
```diff
@@ -1,6 +1,4 @@
 export const Time = {
     NOTIFICATION_DURATION: 4000,
     CURRENT_YEAR: new Date().getFullYear(),
-    START_DATE: new Date(`${new Date().getFullYear()}-01-01`),
-    END_DATE: new Date(`${new Date().getFullYear()}-12-31`)
```
adsukharev, line 5: Don't use it here. It is a potential bug. If user opens app at 31/12/2024 23:59 but after 2 minutes goes to home page, these dates will be old.

frontend/src/types/transactions.ts
```diff
@@ -1,19 +1,16 @@
...
+export interface IOutcome {
+    total: number,
+    transactionGroups: ITransactionGroup[],
 }
 
-export interface IIncomeData {
-    total: number
-}
 
-export interface ITransactionData {
```
adsukharev, line 16: You add everywhere `Data`. What is the value of it?

PR #101, opened 2024-05-09
[TSK-273] add merchant email to Qr-code and send notification based o…

Hoocki:
…n checkbox


adsukharev (approved):

frontend/src/pages/main/payment/components/confirm/PaymentConfirmation.tsx
```diff
@@ -76,9 +74,8 @@ const PaymentConfirmation: React.FC<ConfirmProps> = ({qrCodeData}) => {
             <CardContent className="card-confirm-content">
                 <TextContent text={Content.INFORMATION}/>
                 <BalanceField/>
```
adsukharev: You are getting wallet in PaymentConfirmation and in balance field. On the same page 2 same requests. Bad.
Instead of `<BalanceField/>` use `<FieldConfirmation>` and pass balance and prefix which you have

frontend/src/components/fields/fieldConfirmation/FieldConfirmation.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import {FieldConfirmationProps} from "./props";
+import {InputAdornment, TextField} from "@mui/material";
+import {DISABLE_FIELD} from "../../../common/constansts/fields";
+
+const FieldConfirmation: React.FC<FieldConfirmationProps> = ({data, prefix}) => {
```
adsukharev: bad name. This component has nothing about confirmation. This is just text field which is disabled all the time and has some prefix for value

adsukharev:
Create 2 components: CancelationButton, ConfirmationButton.
Both have required param `handleClick` and optional `buttonName`. Default names for sure: Cancel and Confirm

adsukharev:
Also I found very strange thing. It is in `/common/fields/amountGenerator/`.
1) Looks like files there should generate something. But this is just field folder, how field can generate anything? I think you failed with the name of folder.
2) What is more interesting, there's a component `AmountFieldQrGeneration` . I expect this component should generate Qr code by providing some money, but this is just field with no logic. Why do you name very simple component so complicated if there's no logic about qr code/generation?
3) It is in `common` folder, I expect you use it in some places. But you use it only in one place. If it is, why cannot you move it close to where you use it?

adsukharev:
You have Title and TextContent which are very similar and use the same component Typography. Move it to the same common folder `typography` as you did for `fields`

adsukharev:
I hope you have a good structure in your mind. But for me, it's hard to find/navigate sometimes.. Especially, when I go to constants and there're huge amounts of files. For example, file `positions` what is it? Why you keep it in common/constants if this is kind of css for some header components. Or `authContent` and `content` why did you decide to extract exactly for auth pages some text? Or `fields` has only one interesting constant DISABLE_FIELD, for what ?

frontend/src/pages/main/payment/components/confirm/PaymentConfirmation.tsx
```diff
@@ -76,9 +74,8 @@ const PaymentConfirmation: React.FC<ConfirmProps> = ({qrCodeData}) => {
             <CardContent className="card-confirm-content">
                 <TextContent text={Content.INFORMATION}/>
```
adsukharev: Since this is a actual title, I think you need to use bigger font here (h6) and rename to title smth (maybe CardTitle). But ordinary titles make a bit smaller (h4) and rename to PageTitle

adsukharev: I see you use this title for both Card titles and just text. It's incorrect. Create separate component with h6 for Card titles like here and Transactions in chart

PR #105, opened 2024-05-12
[TSK-278] add Top up page and connection with server

Hoocki:


frontend/src/pages/main/topup/confirm/TopUpConfirmation.tsx
```diff
@@ -0,0 +1,108 @@
...
+        if (!receivedWallet) return;
+        dispatch(showNotification({show: true, type: NotificationType.SUCCESS, message:Notifications.SUCCESSFUL_TOP_UP}));
+        navigate(Paths.ANY);
+    }
+
+    return (
+        <Card className="card-background card-payment">
+            <CardContent className="card-confirm-content">
+                <CardTitle title={Titles.CARD_INFORMATION}/>
+                <CardInformationField cardFieldType={CardFieldType.CARD_NAME} updateField={updateCardField} placeHolder={TopUpPlaceholder.CARD_NAME} isFullWidth={true}/>
```
adsukharev: To make it generic, you made additional type, lots of ifs, switch and so on. I don't think it's a good idea.  It's hard to understand and expand without writing addition code. For me, it's better to just create TextInputs. It's concise and simple. Your solution complicated and even now not generic.

frontend/src/pages/main/topup/confirm/TopUpConfirmation.tsx
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const initialWalletTopUp: IWalletTopUp = {
+    amount: 0
+}
+
+const TopUpConfirmation: React.FC<TopUpConfirmationProps> = ({walletId}) => {
+
+    const [cardData, setCardData] = useState<ICard>(initialCardData);
+    const [sendNotification, setSendNotification] = useState<boolean>(false);
```
adsukharev: why don't you use it?
If you don't have such functionality on the backend side, just remove it from here

frontend/src/pages/main/topup/TopUp.tsx
```diff
@@ -1,11 +1,41 @@
...
+    userId: -1
+}
 
 const TopUp: React.FC = () => {
+
+    const [wallet, setWallet] = useState<IWallet>(initialWallet);
+    const loggedUser = useAppSelector(selectLoggedUser);
+
+    const getWallet = async () => {
+        const receivedWallet = await getWalletByUserService(loggedUser.id, loggedUser.userType);
```
adsukharev: It's time to think about it. You get wallet id on every single page. Can you fetch it once, the reuse it everywhere?

frontend/src/common/constansts/placeholders.ts
```diff
@@ -0,0 +1,6 @@
+export enum TopUpPlaceholder {
```
adsukharev: Why did you create placeholder specifically for top up ?

frontend/src/pages/main/topup/TopUp.tsx
```diff
@@ -1,11 +1,41 @@
...
+
+const initialWallet: IWallet = {
+    id: -1,
+    balance: 0,
+    userId: -1
+}
 
 const TopUp: React.FC = () => {
+
+    const [wallet, setWallet] = useState<IWallet>(initialWallet);
```
adsukharev: What is the advantage of getting wallet here and not in TopUpConfirmation? Same questions for transactions. Why do you get wallet id in Transcations components and not in TableTranscations

frontend/src/pages/main/topup/confirm/TopUpConfirmation.tsx
```diff
@@ -0,0 +1,108 @@
...
+            case CardFieldType.CARD_CCV:
+                setCardData({...cardData, ccv: data as number});
+                break;
+            default:
+                break;
+        }
+
+    }
+
+    const handleNotification = () => {
```
adsukharev:
```suggestion
    const handleSendNotificationCheckbox = () => {
```
Otherwise, it looks like you handle actual notification

frontend/src/pages/main/topup/confirm/TopUpConfirmation.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    return (
+        <Card className="card-background card-payment">
+            <CardContent className="card-confirm-content">
+                <CardTitle title={Titles.CARD_INFORMATION}/>
+                <CardInformationField cardFieldType={CardFieldType.CARD_NAME} updateField={updateCardField} placeHolder={TopUpPlaceholder.CARD_NAME} isFullWidth={true}/>
+                <CardInformationField cardFieldType={CardFieldType.CARD_NUMBER} updateField={updateCardField} placeHolder={TopUpPlaceholder.CARD_NUMBER} isFullWidth={true}/>
+                <Box className="privacy-data-card">
+                    <CardInformationField cardFieldType={CardFieldType.CARD_EXPIRATION_DATE} updateField={updateCardField} placeHolder={TopUpPlaceholder.EXPIRATION_DATE} isFullWidth={false}/>
```
adsukharev: Maybe it can be useful. At least, it's easier in terms of validation, since you don't need to work with string
https://mui.com/x/react-date-pickers/date-picker/#views
https://stackoverflow.com/questions/64112889/month-and-year-picker-in-material-ui-react

frontend/src/pages/main/topup/TopUp.tsx
Hoocki: in Transcations i faced with loop requests to the server when i tried to get wallet from server, so i moved to the parent component call `getWallet`. now i get `walletId` from redux, so it was removed.

frontend/src/stores/redux/wallet/walletSlicer.ts
```diff
@@ -0,0 +1,26 @@
...
+    initialState,
+    reducers: {
+        addWallet: (state, action: PayloadAction<IWallet>) => ({...action.payload}),
+        clearWallet: state => ({...initialState})
+    }
+})
+
+export const {addWallet, clearWallet} = walletSlice.actions;
+
+export const selectWalletId = (state: RootState) => state.walletState.id;
```
adsukharev: if you use only walletId, don't store other info

adsukharev (approved):

PR #106, opened 2024-05-24
[TSK-291] add transactions chart

Hoocki:


frontend/src/common/constansts/paths.ts
```diff
@@ -2,7 +2,7 @@ export enum Paths {
     PROFILE = "/profile",
     TOP_UP = "/topup",
     TRANSACTIONS = "/transactions",
-    HOME = "",
+    HOME = "/",
```
adsukharev, line 5: Don't forget to check it in the next ticket

transaction-history-manager/src/main/java/com/qpay/transactionhistorymanager/service/TransactionHistoryService.java
```diff
@@ -2,14 +2,18 @@
...
 import java.util.List;
 
 public interface TransactionHistoryService {
 
     List<TransactionEntity> getPageOfTransactionsByUserId(long userId, final Pageable pageable, final UserType userType);
 
     TransactionEntity saveTransaction(TransactionModification transactionModification);
-}
+
+    TransactionOutcome getTransactionsOutcome(long walletId, LocalDateTime startDateTime, LocalDateTime endDateTime);
```
adsukharev, line 18: When you work on app which is world wide used, you cannot use localDateTime, because every user lives in a different time zone. In that case you should use ZoneDateTime or Instant

transaction-history-manager/src/main/java/com/qpay/transactionhistorymanager/service/impl/TransactionHistoryServiceImpl.java
```diff
@@ -31,4 +47,112 @@ public TransactionEntity saveTransaction(final TransactionModification transacti
...
+                filterTransactions,
+                groupByIdTo,
+                sortDesc,
+                groupTransactionsAndCalculateTotal,
+                limitTransactionsGroup,
+                unwindTransactionsGroup,
+                addShareField,
+                finalGroupOperation,
+                finalProjection
+        );
```
adsukharev: I would create a dedicated class, e.g. `AggregationPipelineGenerator` which returns this pipeline `final var aggregation = Aggregation.newAggregation(...`

adsukharev (approved):

PR #107, opened 2024-06-08
[TSK-259] sync updated design, improvements

kostya4201337:


frontend/src/components/header/Header.tsx
```diff
@@ -64,14 +65,13 @@ const Header: React.FC<HeaderProps> = ({isLogged}: HeaderProps) => {
     }, [loggedUserType]);
 
     return (
-        <AppBar position="static" className="app-bar">
+        <AppBar className="app-bar">
             {isLogged && (
-                <Container maxWidth="xl">
+                <Container maxWidth="lg">
```
Hoocki: if you can move it to css it would be good

frontend/src/components/header/desktopMenu/styles.css
```diff
@@ -15,5 +15,6 @@
     margin-left: 3rem;
     font-size: 1rem;
     font-family: sans-serif;
+    font-weight: 300;
```
Hoocki, line 18: 300 of what? px, rem?

frontend/src/components/header/styles.css
```diff
@@ -1,3 +1,10 @@
 .app-bar {
     background-color: white;
+    position: absolute;
+    margin: 0;
+    padding: 0;
```
Hoocki, line 5: same question

frontend/src/components/header/userMenu/confirmLogOut/styles.css
```diff
@@ -0,0 +1,16 @@
+.title {
+    font-size: 2.5rem;
+    font-weight: bold;
+}
+
+.content {
+    font-weight: 200;
```
Hoocki, line 7: same question

frontend/src/components/header/userMenu/styles.css
```diff
@@ -19,4 +19,5 @@
     color: black;
     text-align: center;
     margin-right: 1rem;
+    font-weight: 300;
```
Hoocki, line 22: ok, i understand that you don't add parameter, so check every single css and add it

frontend/src/components/typography/cardTitle/styles.css
Hoocki, line 1: delete it, if you don't use it

Hoocki (commented):
1) in all pages i can scroll down, why?
2) in Sign up page, footer overlaps
3) in Transactions, why i can click to the next page of transactions, if in current page their fewer than max pages size?

frontend/src/components/header/desktopMenu/styles.css
konstantinwas: image

frontend/src/components/header/styles.css
konstantinwas: image

frontend/src/components/header/Header.tsx
konstantinwas: no, i can move it to styles.ts but why

Hoocki: styles should be in css so that they do not interfere with reading the code

konstantinwas:
> 3\. in Transactions, why i can click to the next page of transactions, if in current page their fewer than max pages size?

 because when mock enabled, it return same transactions regardless of the page

konstantinwas:
> * in Sign up page, footer overlaps

what do you mean

Hoocki:
> > 3. in Transactions, why i can click to the next page of transactions, if in current page their fewer than max pages size?
> 
> because when mock enabled, it return same transactions regardless of the page

i tested it with backend and i can click to the next page when transactions fewer than max pages

Hoocki:
> > * in Sign up page, footer overlaps
> 
> what do you mean

run client and go ahead to the Sign up page. You will see footer overlaps

Hoocki (approved):

konstantinwas:
@adsukharev

frontend/src/common/themes/button.ts
```diff
@@ -0,0 +1,37 @@
...
+declare module '@mui/material/Button' {
+    interface ButtonPropsColorOverrides {
+        blue: true;
+        cancel: true;
+    }
+}
+
+export const buttonTheme = createTheme({
+    palette: {
+        blue: {
```
adsukharev: blue is a bad name. Maybe primary?

frontend/src/common/themes/button.ts
```diff
@@ -0,0 +1,37 @@
+import {createTheme} from "@mui/material/styles";
+
+declare module '@mui/material/styles' {
+    interface Palette {
+        blue: Palette['primary'];
+        cancel: Palette['primary'];
```
adsukharev: why blue and cancel are both relates to primary?

frontend/src/components/buttons/cancelButton/CancelButton.tsx
```diff
@@ -1,18 +1,21 @@
...
     return (
-        <Button
-            onClick={handleClick}
-            variant="contained"
-            className="button"
-            color="secondary"
-        >
-            {buttonName ? buttonName : Buttons.CANCEL}
-        </Button>
+        <ThemeProvider theme={buttonTheme}>
```
adsukharev: to not write <ThemeProvider evrywhere, move it to Providers component to apply it once and for all

frontend/src/common/themes/button.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+declare module '@mui/material/Button' {
+    interface ButtonPropsColorOverrides {
+        blue: true;
+        cancel: true;
+    }
+}
+
+export const buttonTheme = createTheme({
```
adsukharev: I expect that these colors are for all components, not only for buttons.

frontend/src/common/styles/container.ts
```diff
@@ -0,0 +1,6 @@
+export const cardAuthSx = {
```
adsukharev: why have you created style specifically for sx?

frontend/src/components/buttons/confirmationButton/ConfirmationButton.tsx
```diff
@@ -1,18 +1,22 @@
...
+        <ThemeProvider theme={buttonTheme}>
+            <Button
+                onClick={handleClick}
+                variant="contained"
+                color="blue"
+                className={`button ${className}`}
+                disabled={isDisabled}
+            >
+                {buttonName ? buttonName : Buttons.CONFIRM}
+            </Button>
```
adsukharev: It would be best if you don't need to touch this and cancel button components at all. It means you rewrite primary and secondary color styles and they automatically appplies for all components in your project where you have already used these styles

konstantinwas: But i already don't have to touch these components when i change primary or secondary color? i dont understand what you mean

adsukharev: add changes to styles such way that you don't need to change this file at all

